Issue #9, opened 2018-01-31
Promote dspec.py to collaboration level


jsdillon:
Generalize the code but also preserved backwards compatibility for capo/PAPER users as much as possible.

PR #10, opened 2018-01-31
Update delay filter

jsdillon:
This PR addresses #9 and updates the delay filtering functionality in dspec.py to have much clearer documentation and be more generally useful for other kinds of filtering problems.

This also makes it so that travis can run, though it errors on other modules in uvtools (not our problem right now, they still need to be promoted).


uvtools/dspec.py
```diff
@@ -1,78 +1,160 @@
...
+    return calc_width(bl_dly, sdf, nchan)    
+
+
+def calc_width(filter_size, real_delta, nsamples):
+    '''Calculate the upper and lower bin indices of a fourier filter.
+
+    Arguments:
+        filter_size: the half-width (i.e. the width of the positive part) of the region in fourier 
+            space, symmetric about 0, that is filtered out. In units of 1/[real_delta].
+        real_delta: the bin width in real space
```
philbull: Does this need units?

uvtools/dspec.py
```diff
@@ -1,78 +1,160 @@
...
+        wgts: real numpy array of multiplicative weights with the same shape as the data. 
+        filter_size: the half-width (i.e. the width of the positive part) of the region in fourier 
+            space, symmetric about 0, that is filtered out. In units of 1/[real_delta].
+        real_delta: the bin width in real space of the dimension to be filtered
+        tol: CLEAN algorithm convergence tolerance (see aipy.deconv.clean)
+        window: window function for filtering applied to the filtered axis. 
+            See aipy.dsp.gen_window for options.
+        skip_wgt: skips filtering highly flagged rows (unflagged fraction ~< skip_wgt) 
+            Only works properly when all weights are all between 0 and 1.
+        maxiter: Maxinum number of iterations for aipy.deconv.clean to converge.
```
philbull: Typo

uvtools/dspec.py
```diff
@@ -1,78 +1,160 @@
...
-    '''Apply a wideband delay filter to data.  Data are weighted again by wgts,
-    windowed, Fourier transformed, and deconvolved allowing clean components
-    between lthresh and uthresh.  The mdl, residual, and info are returned in
-    frequency domain.'''
+    if lthresh == 0: 
+        lthresh = nsamples
+    return (uthresh, lthresh)
+
+
+def high_pass_fourier_filter(data, wgts, filter_size, real_delta, tol=1e-9, window='none', skip_wgt=0.1, maxiter=100):
```
philbull: This seems to return the high-pass and low-pass components, so could the function name be simplified/generalised?

uvtools/dspec.py
```diff
@@ -1,78 +1,160 @@
...
+    Arguments:
+        data: 1D or 2D (real or complex) numpy array to be filtered along the last dimension.
+        wgts: real numpy array of multiplicative weights with the same shape as the data. 
+        filter_size: the half-width (i.e. the width of the positive part) of the region in fourier 
+            space, symmetric about 0, that is filtered out. In units of 1/[real_delta].
+        real_delta: the bin width in real space of the dimension to be filtered
+        tol: CLEAN algorithm convergence tolerance (see aipy.deconv.clean)
+        window: window function for filtering applied to the filtered axis. 
+            See aipy.dsp.gen_window for options.
+        skip_wgt: skips filtering highly flagged rows (unflagged fraction ~< skip_wgt)
```
philbull: "Highly flagged" means the value of `wgts` is low, right?

uvtools/dspec.py
```diff
@@ -1,78 +1,160 @@
...
-        data = binning.phs2lstbin(data, aa, i, j, jds=jds, lst_res=lst_res)
-    bl = aa.get_baseline(i,j)
-    return delay_filter(data, wgts, np.linalg.norm(bl), sdf, 
-            standoff=standoff, horizon=horizon, tol=tol, window=window, 
-            skip_wgt=skip_wgt, maxiter=maxiter)
+
+    
+def delay_filter(data, wgts, bl_len, sdf, standoff=0., horizon=1., tol=1e-4, 
+        window='none', skip_wgt=0.5, maxiter=100):
+    '''Apply a wideband delay filter to data. Variable names preserved for
```
philbull: The function names have changed, so do we need to keep the old variable names? If backwards compatibility is important, it might be worth writing a thin wrapper around the new functions that provides functions with the old names and keywords.

uvtools/dspec.py
```diff
@@ -1,78 +1,160 @@
...
+    bl_dly = horizon * bl_len + standoff
+    return high_pass_fourier_filter(data, wgts, bl_dly, sdf, tol=tol,window=window,
+                                    skip_wgt=skip_wgt, maxiter=maxiter)
+
+
+
+
+#import binning
+
+# def delay_filter_aa(aa, data, wgts, i, j, sdf, phs2lst=False, jds=None,
```
philbull: Can this commented-out code just be removed?

uvtools/dspec.py
jsdillon: No. It's general. The filter size has units that depend on it's units though.

uvtools/dspec.py
jsdillon: That's totally fair, but to maintain consistency with how it's been used and how delay_filter works, I think we can continue to call it the high-pass filter since the "main" product is the residual...even though it's the 2nd argument returned.

uvtools/dspec.py
jsdillon: Clarified.

uvtools/dspec.py
jsdillon: Fixed.

uvtools/dspec.py
jsdillon: Probably, but it's a useful reminder to anyone looking for old code that these functions used to exist (but I've commented them out since they aren't being unit tested). If in a year no one is clamoring for them, we can delete them.

uvtools/dspec.py
jsdillon: delay_filter's function name actually has not changed... though github's diff makes that a bit hard to see.

uvtools/dspec.py
```diff
@@ -1,78 +1,160 @@
...
+    
+    Arguments:
+        bl_len: length of baseline (in 1/[sdf], typically ns)
+        sdf: frequency channel width (typically in GHz)
+        nchan: number of frequency channels
+        standoff: fixed additional delay beyond the horizon (same units as bl_len)
+        horizon: proportionality constant for bl_len where 1 is the horizon (full light travel time)
+
+    Returns:
+        uthresh, lthresh: bin indices for upper and lower edges of the fourier filter
```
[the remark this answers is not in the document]
jsdillon: Good question... let me look into it.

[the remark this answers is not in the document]
jsdillon: It's a bit tricky. The uthresh is positive (counts from front) and is filtered. The lthresh is negative (counts from back) and is not filtered. I updated the documentation to clarify.

PR #21, opened 2018-05-01
added win_kwargs to pass through to aipy.dsp.gen_window

nkern:
so that we can use the `tukey` window

### DIFF
--- a/uvtools/dspec.py
+++ b/uvtools/dspec.py
@@ -43,7 +43,7 @@ def calc_width(filter_size, real_delta, nsamples):
     return (uthresh, lthresh)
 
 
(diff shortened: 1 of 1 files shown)


uvtools/dspec.py
```diff
@@ -61,14 +61,15 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, tol=1e-9, wind
             Model is left as 0s, residual is left as data, and info is {'skipped': True} for that 
             time. Only works properly when all weights are all between 0 and 1.
         maxiter: Maximum number of iterations for aipy.deconv.clean to converge.
+        win_kwargs : keyword arguments for window function selection (see aipy.dsp.gen_window)
```
zakiali: Could you put some more information for the win_kwargs argument? Is it an integer? float? ranges? It's a little annoying to go dig in aipy for what this is.

PR #22, opened 2018-06-01
added frf.py and FRFilter class with tests

nkern:
Adds `frf.py` for Fringe Rate Filtering capabilities. In `frf.py`, however, there doesn't exist a traditional FRF (as is done in `fringe.py`), but currently only has simpler time averaging functions. At some point we should move over the more traditional fringe rate filtering code into the object oriented structure in `frf.py`. These basic capabilities provided here are what we need in the short-term for IDR2.1 analysis


nkern:
This code is better suited to be in `hera_cal`. so that's where its going...

PR #23, opened 2018-06-02
added utils.py and utils.search_data with tests

nkern:
Adds a simple function to look for data file templates, (e.g. `zen.even.{pol}.LST.*.uv`)

### DIFF
--- a/uvtools/utils.py
+++ b/uvtools/utils.py
@@ -1,0 +1,82 @@
+import numpy as np
+import os
+import glob
+
+
+def search_data(templates, pols, matched_pols=False, reverse_nesting=False):
+    """
+    Glob-parse data templates to search for data files.
+
+    Parameters
+    ---------- 
+    templates : str or list
+        A glob-parsable search string, or list of such strings, with a {pol}
+        spot for string formatting. Ex. ["zen.even.{pol}.LST.*.HH.uv"]
+
+    pols : str or list
+        A polarization string, or list of polarization strings, to search for.
+        Ex. ["xx", "yy"]
+
+    matched_pols : boolean
+        If True, only use datafiles that are present for all polarizations.
+
+    reverse_nesting : boolean
+        If True, flip the nesting of datafiles to be polarization-datafile.
+        By default, the output is polarization-datafile.
+
+    Returns
+    -------
+    datafiles : list
+        A nested list of paths to datafiles. By default, the structure is
+        polarization-datafile nesting. If reverse_nesting, then the structure
+        is flipped to datafile-polarization structure.
+    """
+    # type check
+    if isinstance(templates, (str, np.str)):
+        templates = [templates]
+    if isinstance(pols, (str, np.str)):
+        pols = [pols]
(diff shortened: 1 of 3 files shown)


nkern:
sure thing. thanks @plaplant for the catch

uvtools/utils.py
```diff
@@ -0,0 +1,101 @@
...
+        _templates = []
+        for f in unique_files:
+            goodfile = True
+            for p in pols:
+                if f.format(pol=p) not in allfiles:
+                    goodfile = False
+            if goodfile:
+                _templates.append(f)
+
+        datafiles, datapols = search_data(_templates, pols, matched_pols=False, reverse_nesting=False)
```
plaplant: You might want to add a comment explaining why you're recursively calling the function here (which I think is fine, it's just nice to know why).

uvtools/utils.py
```diff
@@ -0,0 +1,101 @@
...
+    # reverse nesting if desired
+    if reverse_nesting:
+        datafiles = []
```
plaplant: Re-instantiating these variables as empty lists here is a little counter intuitive. Seems like you're undoing all the work you did above to build them. Perhaps this work to invert the nesting can be done as you're building the list?

nkern:
Thanks @plaplant for the comments.

uvtools/utils.py
nkern: Yeah, it seems like i'm undoing the previous work, but the information from before is preserved in the `allfiles` list, which is needed in this loop and was constructed from the first loop

plaplant: So if this is the case, why can't we do the reversal when initially constructing the lists?

nkern: `allfiles` isn't constructed until after the first loop over `pols` and `templates` is finished

plaplant: Right, but does that mean there can be repeated items when `reverse_nesting==False`? Regardless, it seems like the function can be reorganized, because if `matched_pols`, `reverse_nesting`, and `flatten` are all `False`, then the function is doing unnecessary work to generate the `unique_files` list which is then never used.

nkern: I don't think there can be repeated items when `reverse_nesting==False`. The key difference in the second time around (when `reverse_nesting` == True) is that the search templates (in `unique_files`) no longer have wildcard characters in them, so a `glob.glob` yields a single file (hence the variable `df` not `dfs`).

Yeah there is a ***little*** bit of extra computation done if all the kwargs are set to False (as they are by default), but its just list comprehensions and string formatting, which is fairly fast, and is essentially negligible overhead. If you'd like I can add check kwarg and `return`statement before those calculations are done, but I think its unnecessary given how little time it takes anyways.

plaplant: Ah, didn't realize that the wildcards had been expanded.

Okay, fair enough. It's true there's not much overhead incurred, it's more the principle of doing unnecessary work than the work itself. But it's fine, I'll approve and merge as is.

PR #26, opened 2018-07-17
Add new functions to do leastsq fits of Fourier series to flagged data

philbull:
This PR implements a simple Fourier series fitting algorithm for flagged data. This can be used to in-paint flagged data, or as an alternative to CLEAN deconvolution. Because of the way the fitting is done, it largely avoids introducing ringing and edge effects (unlike FFT).


uvtools/dspec.py
```diff
@@ -146,6 +147,250 @@ def delay_filter(data, wgts, bl_len, sdf, standoff=0., horizon=1., min_dly=0.0,
...
+        
+    model_coeffs : array_like
+        Coefficients of Fourier modes, ordered from modes [-n, +n].
+    
+    data_out : array_like
+        In-painted data.
+    """
+    # Construct Fourier basis operator if not specified
+    if operator is None:
+        # Frequency array
```
plaplant: This block of code is used in the function above too. It might make sense to make the basis generation its own function, to make sure it will be the same for both uses. (I'm assuming that the average user will call this function without specifying their own basis, and then want the option to use `fourier_model` to generate an actual spectrum from just the saved coefficients. If you had a different use case in mind, then this comment might be less relevant.)

uvtools/dspec.py
```diff
@@ -146,6 +147,250 @@ def delay_filter(data, wgts, bl_len, sdf, standoff=0., horizon=1., min_dly=0.0,
...
+        nu = np.arange(data.size)
+        L = nu[-1] - nu[0]
+        
+        # Build matrix operator for complex Fourier basis
+        n = np.arange(-nmax, nmax+1)
+        F = np.array([np.exp(-1.j * _n * nu / L) for _n in n])
+    else:
+        F = operator
+        n = np.arange(-nmax, nmax+1)
+        assert F.shape[0] == 2*nmax+1, "Fourier basis operator has the wrong shape"
```
plaplant: In this case, I feel more strongly that this should be in an actual `if...raise` block, since the user is explicitly providing their own operator that needs to be the right shape.

uvtools/dspec.py
```diff
@@ -146,6 +147,250 @@ def delay_filter(data, wgts, bl_len, sdf, standoff=0., horizon=1., min_dly=0.0,
...
+    
+    # Turn flags into a mask
+    w = np.logical_not(flags)
+    
+    # Define model and likelihood function
+    model = lambda cn: np.dot(cn, F)
+    
+    def loglike(cn):
```
plaplant: Since this is being passed in as the objective function for `leastsq`, it should probably be defined down near line 290, in the `else` block of `use_linear`. It should also have a brief docstring of its own.

uvtools/dspec.py
```diff
@@ -146,6 +147,250 @@ def delay_filter(data, wgts, bl_len, sdf, standoff=0., horizon=1., min_dly=0.0,
...
+        # Need to do real and imaginary parts separately, otherwise leastsq() fails
+        _delta = w * (data - model(cn[:n.size] + 1.j*cn[n.size:]))
+        delta = np.concatenate((_delta.real/sigma, _delta.imag/sigma))
+        return -0.5 * delta**2.
```
plaplant: Can you provide a brief comment/reference for why you're returning this quantity? Given the name of the function (`loglike`), it's not immediately obvious (to me at least).

uvtools/dspec.py
```diff
@@ -146,6 +147,250 @@ def delay_filter(data, wgts, bl_len, sdf, standoff=0., horizon=1., min_dly=0.0,
...
+    
+    # Initial guess for Fourier coefficients (real + imaginary blocks)
+    cn_in = np.zeros(2*n.size)
+    if cn_guess is not None:
+        assert cn_in.size == 2*cn_guess.size, "cn_guess must be of size %s" \
```
plaplant: Like `operator`, this should probably have a more stringent test, since the user is providing this explicitly.

uvtools/dspec.py
```diff
@@ -146,6 +147,250 @@ def delay_filter(data, wgts, bl_len, sdf, standoff=0., horizon=1., min_dly=0.0,
...
+            data_out[flags] += sigma * noise
+    
+    # Return coefficients and best-fit model
+    return bf_model, cn_out, data_out
+
+
+def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, 
+                         cn_guess=None, use_linear=True):
+    """
+    Fit a smooth model to 2D complex-valued data with flags, using a linear
```
plaplant: I think it would be useful to mention that under the hood this just calls `delay_filter_leastsq_1d` along the first axis of the `data` array (i.e., it's not doing a two-dimensional fit).

uvtools/tests/test_dspec.py
```diff
@@ -75,6 +75,55 @@ def test_delay_filter_2D(self):
         np.testing.assert_allclose(np.average(data,axis=1), np.average(dmdl,axis=1), atol=1e-3)
         np.testing.assert_allclose(np.average(dres,axis=1), 0, atol=1e-3)
     
+    def test_delay_filter_leastsq(self):
```
plaplant: It would be good to have explicit tests for `fourier_model`, and to break the tests for the 1d and 2d versions into different functions.

uvtools/tests/test_dspec.py
```diff
@@ -75,6 +75,55 @@ def test_delay_filter_2D(self):
...
+        cn = np.array([-0.1-0.1j, -0.1+0.1j, -0.3-0.01j, 
+                        0.5+0.01j, 
+                       -0.3-0.01j, -0.1+0.1j, 0.1-0.1j])
+        data *= np.atleast_2d( dspec.fourier_model(cn, NCHAN) )
+        
+        # Estimate smooth Fourier model on unflagged data
+        bf_model, cn_out, data_out = dspec.delay_filter_leastsq(data, flags, 
+                                                                sigma, nmax=3, 
+                                                                add_noise=False)
+        np.testing.assert_allclose(data.real, bf_model.real, atol=NCHAN*TOL)
```
plaplant: I know it's pedantic, but could we add tests that the imaginary values are close too?

uvtools/tests/test_dspec.py
```diff
@@ -75,6 +75,55 @@ def test_delay_filter_2D(self):
...
+    def test_delay_filter_leastsq(self):
+        NCHAN = 128
+        NTIMES = 10
+        TOL = 1e-7
+        data = np.ones((NTIMES, NCHAN), dtype=np.complex)
+        flags = np.zeros((NTIMES, NCHAN), dtype=np.bool)
+        sigma = 0.1 # Noise level (not important here)
+        
+        # Fourier coeffs for input data, ordered from (-nmax, nmax)
+        cn = np.array([-0.1-0.1j, -0.1+0.1j, -0.3-0.01j,
```
plaplant: Is it useful to see that the solved Fourier coefficients are comparable to the input ones? Or do we not expect them to be the same?

uvtools/tests/test_dspec.py
```diff
@@ -75,6 +75,55 @@ def test_delay_filter_2D(self):
...
+        
+        # Test 1D code with non-linear leastsq
+        bf_model, cn_out, data_out = dspec.delay_filter_leastsq_1d(
+                                                    data[0], flags[0], sigma, 
+                                                    nmax=3, add_noise=False, 
+                                                    use_linear=False)
+        np.testing.assert_allclose(data[0], bf_model, atol=NCHAN*TOL)
+        
+        # Test that noise injection can be switched on
+        bf_model, cn_out, data_out = dspec.delay_filter_leastsq_1d(
```
plaplant: There should be an explicit check of the output here, to make sure the code is producing expected values, rather than just running without erroring.

uvtools/dspec.py
```diff
@@ -146,6 +147,250 @@ def delay_filter(data, wgts, bl_len, sdf, standoff=0., horizon=1., min_dly=0.0,
...
+        Initial guess for the series coefficients. If None, zeros will be used. 
+        A sensible choice of cn_guess can speed up the solver significantly. 
+        Default: None.
+    
+    use_linear : bool, optional
+        Whether to use a fast linear least-squares solver to fit the Fourier 
+        coefficients, or a slower generalized least-squares solver. 
+        Default: True.
+    
+    operator : array_like, optional
```
plaplant: Because this argument is optional, you should mention what the default value is. ~~Also, do we want to have the option of letting the user specify the basis? Put another way, are there well-motivated reasons to _not_ use the regular Fourier basis? It seems like it could get confusing to keep track of when trying to reconstruct things later, especially since the produced coefficients depend on the basis passed in.~~ Disregard this, see comment below on the 2d version.

uvtools/dspec.py
```diff
@@ -146,6 +147,250 @@ def delay_filter(data, wgts, bl_len, sdf, standoff=0., horizon=1., min_dly=0.0,
...
+    mdl_array = np.zeros(data.shape, dtype=np.complex)
+    
+    # Loop over array
+    cn_out = None
+    for i in range(data.shape[0]):
+        bf_model, cn_out, data_out = delay_filter_leastsq_1d(
+                                                data[i], flags[i], sigma=sigma, 
+                                                nmax=nmax, add_noise=add_noise, 
+                                                use_linear=use_linear,
+                                                cn_guess=cn_out, operator=F)
```
plaplant: Ah, I see, the `operator` argument is being pre-computed to save on computation for the 2d version. In that case, we should explicitly say that in the docs, and recommend against the user providing their own arbitrary basis.

uvtools/dspec.py
philbull: Done!

uvtools/dspec.py
philbull: Done.

uvtools/dspec.py
philbull: It's just a simple Gaussian log-likelihood. Comments added.

uvtools/dspec.py
philbull: Done!

uvtools/dspec.py
philbull: Done.

uvtools/dspec.py
philbull: Done.

uvtools/dspec.py
philbull: Done.

uvtools/tests/test_dspec.py
philbull: I've done the first part, but kept the 1D and 2D tests together, as I don't think there's much to be gained by separating them.

uvtools/tests/test_dspec.py
philbull: Added a couple of tests to do this in the unmasked case.

uvtools/tests/test_dspec.py
philbull: See above; added tests to check cn values.

uvtools/tests/test_dspec.py
philbull: Done.

PR #30, opened 2018-12-19
updated dspec.high_pass* handling of residual term

nkern:
Updates how `dspec.high_pass_fourier_filter` handles the residual term. It makes adding the residual within the CLEAN window to the model optional, and uses the residual from `aipy.deconv.clean`, rather than re-compute it at the end as `data - model`. This was done because for a pure filtering process, one really should be adding the residual within the filtering window to the model, but b/c this module does both CLEANing and filtering, I wanted to make it optional.

The function now also outputs the "CLEANed" data, which is `mdl + res`.

### DIFF
--- a/uvtools/dspec.py
+++ b/uvtools/dspec.py
@@ -162,7 +162,7 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False,
                     info.append({'skipped': True})
                 else:
                     _d_cl, info_here = aipy.deconv.clean(_d[i], _w[i], area=area, tol=tol, stop_if_div=False, maxiter=maxiter, gain=gain)
-                    d_mdl[i] = np.fft.fft(_d_cl + info_here['res'] * area)
+                    d_mdl[i] = np.fft.fft(_d_cl)
                     del info_here['res']
                     info.append(info_here)
 
@@ -197,7 +197,7 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False,
             
         # run clean
         _d_cl, info = aipy.deconv.clean(_d, _w, area=area, tol=tol, stop_if_div=False, maxiter=maxiter, gain=gain)
-        d_mdl = np.fft.fft2(_d_cl + info['res'] * area, axes=(0, 1))
+        d_mdl = np.fft.fft2(_d_cl, axes=(0, 1))
         del info['res']
 
     d_res = data - d_mdl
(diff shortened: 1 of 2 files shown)


uvtools/dspec.py
```diff
@@ -194,20 +201,34 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False,
...
+    if clean2d:
+        d_mdl = np.fft.fft2(_d_cl, axes=(0, 1))
+        d_res = np.fft.fft2(_d_res, axes=(0, 1))
+    else:
+        d_mdl = np.fft.fft(_d_cl)
+        d_res = np.fft.fft(_d_res)
 
-    return d_mdl, d_res, info
+    # form clean data
+    d_cln = d_mdl + d_res
```
jsdillon: I'd advocate removing this, but if you don't having `_d_cl` FFT to `d_mdl` while you also have a `d_cln` is very confusing. If you remove `d_cln` then it doesn't matter.

uvtools/dspec.py
```diff
@@ -194,20 +201,34 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False,
...
+                               
+    return d_mdl, d_res, d_cln, info
 
 
 def delay_filter(data, wgts, bl_len, sdf, standoff=0., horizon=1., min_dly=0.0, tol=1e-4,
```
jsdillon: Be sure to propagate any documentation changes from `high_pass_fourier_filter` here.

uvtools/dspec.py
```diff
@@ -252,11 +278,12 @@ def delay_filter(data, wgts, bl_len, sdf, standoff=0., horizon=1., min_dly=0.0,
...
     # run fourier filter
     return high_pass_fourier_filter(data, wgts, bl_dly, sdf, tol=tol, window=window, edgecut_low=edgecut_low,
-                                    edgecut_hi=edgecut_hi, skip_wgt=skip_wgt, maxiter=maxiter, gain=gain, alpha=alpha)
+                                    edgecut_hi=edgecut_hi, skip_wgt=skip_wgt, maxiter=maxiter, gain=gain,
+                                    alpha=alpha, add_clean_residual=add_clean_residual)
 
 
-def fringe_filter(data, wgts, max_frate, dt, tol=1e-4, skip_wgt=0.5, maxiter=100,
-                  gain=0.1, window='none', edgecut_low=0, edgecut_hi=0, alpha=0.5):
+def fringe_filter(data, wgts, max_frate, dt, tol=1e-4, skip_wgt=0.5, maxiter=100, gain=0.1,
```
jsdillon: Be sure to propagate any documentation changes from `high_pass_fourier_filter` here.

PR #33, opened 2019-04-18
7term BH window update

nkern:
Adds a 7-term BH window for even more sidelobe suppression than normal blackman-harris window (which in scipy is default 4-term window)

### DIFF
--- a/uvtools/dspec.py
+++ b/uvtools/dspec.py
@@ -457,9 +457,21 @@ def gen_window(window, N, alpha=0.5, edgecut_low=0, edgecut_hi=0, **kwargs):
     elif window == 'tukey':
         w[edgecut_low:edgecut_hi] =  windows.tukey(N - Ncut, alpha)
     elif window in ['blackmanharris-7term', 'blackman-harris-7term', 'bh7']:
-        # https://ieeexplore.ieee.org/document/940309
+        # https://ieeexplore.ieee.org/document/293419
         a_k = [0.27105140069342, -0.43329793923448, 0.21812299954311, -0.06592544638803,
                0.01081174209837, -0.00077658482522, 0.00001388721735]
+        w[edgecut_low:edgecut_hi] = np.sum([[ak * np.cos(2 * np.pi * k * n / float(N - Ncut - 1)) for n in range(N - Ncut)] for k, ak in enumerate(a_k)], axis=0)
+    elif window in ['cosinesum-9term', 'cosinesum9term', 'cs9']:
+        # https://ieeexplore.ieee.org/document/940309
+        a_k = [2.384331152777942e-1, -4.00554534864382e-1, 2.358242530472107e-1, -9.527918858383112e-2,
+               2.537395516617152e-2, -4.152432907505835e-3, 3.68560416329818e-4, -1.38435559391703e-5,
+               1.161808358932861e-7]
+        w[edgecut_low:edgecut_hi] = np.sum([[ak * np.cos(2 * np.pi * k * n / float(N - Ncut - 1)) for n in range(N - Ncut)] for k, ak in enumerate(a_k)], axis=0)
+    elif window in ['cosinesum-11term', 'cosinesum11term', 'cs11']:
+        # https://ieeexplore.ieee.org/document/940309
+        a_k = [2.151527506679809e-1, -3.731348357785249e-1, 2.424243358446660e-1, -1.166907592689211e-1,
+               4.077422105878731e-2, -1.000904500852923e-2, 1.639806917362033e-3, -1.651660820997142e-4,
+               8.884663168541479e-6, -1.938617116029048e-7, 8.482485599330470e-10]
         w[edgecut_low:edgecut_hi] = np.sum([[ak * np.cos(2 * np.pi * k * n / float(N - Ncut - 1)) for n in range(N - Ncut)] for k, ak in enumerate(a_k)], axis=0)
     else:
         try:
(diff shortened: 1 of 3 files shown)


uvtools/dspec.py
```diff
@@ -450,12 +450,17 @@ def gen_window(window, N, alpha=0.5, edgecut_low=0, edgecut_hi=0, **kwargs):
...
-    elif window in ['blackmanharris', 'blackman-harris']:
+    elif window in ['blackmanharris', 'blackman-harris', 'bh', 'bh4']:
         w[edgecut_low:edgecut_hi] =  windows.blackmanharris(N - Ncut)
     elif window in ['hanning', 'hann']:
         w[edgecut_low:edgecut_hi] =  windows.hann(N - Ncut)
     elif window == 'tukey':
         w[edgecut_low:edgecut_hi] =  windows.tukey(N - Ncut, alpha)
+    elif window in ['blackmanharris-7term', 'blackman-harris-7term', 'bh7']:
+        # https://ieeexplore.ieee.org/document/940309
+        a_k = [0.27105140069342, -0.43329793923448, 0.21812299954311, -0.06592544638803,
```
plaplant: The coefficients you hard-code here are (slightly) different than those in the paper you cite. Why are they different? Where are these coefficients coming from?

uvtools/dspec.py
```diff
@@ -450,12 +450,17 @@ def gen_window(window, N, alpha=0.5, edgecut_low=0, edgecut_hi=0, **kwargs):
...
+               0.01081174209837, -0.00077658482522, 0.00001388721735]
+        w[edgecut_low:edgecut_hi] = np.sum([[ak * np.cos(2 * np.pi * k * n / float(N - Ncut - 1)) for n in range(N - Ncut)] for k, ak in enumerate(a_k)], axis=0)
```
plaplant: This line will produce division by zero if `N` is 1 (or if `N - Ncut` is 1). We should check for this case, and return an array with a value of `1.0` (the same behavior as for the scipy Blackman-Harris window).

uvtools/dspec.py
nkern: this was from stack overflow and the paper was a reference. i couldn't open it but figured i'd link in as a comment anyways for transparency. i could just post the stackoverflow thread if you'd prefer

plaplant: I'd actually rather use the values in the paper in this case, because the author explicitly says that computing the coefficients naively can lead to computational precision issues, even with double precision arithmetic. Not knowing the provenance of the stackoverflow coefficients, it seems safer to use the values in the paper.

uvtools/dspec.py
nkern: we do this a few lines above

uvtools/dspec.py
nkern: ok!

nkern: Actually, this paper is not the standard 7-term BH window, but uses a different method for sidelobe minimization and is just a general family of cosine-summed windows. I did find the relevant 7-term BH window paper and the coefficients match what it says, so I'll just link to that instead!

nkern:
@plaplant okay so I linked the correct paper for the 7-term BH, but I also added the 9-term and 11-term cosine-sum windows (which achieve even greater sidelobe suppression) from the original paper I linked. Here is an example plot of their sidelobes for posterity.

Figure_1

uvtools/dspec.py
plaplant: Great, thanks for finding the right link!

uvtools/dspec.py
plaplant: If `N` is 1, and `Ncut` is 0 (as it would be if `edgecut_low` and `edgecut_hi` are not specified), then the ValueError is not raised, and so we get a ZeroDivisionError:
```
In [2]: gen_window("bh7", 1)
---------------------------------------------------------------------------
ZeroDivisionError                         Traceback (most recent call last)
<ipython-input-2-c6dafc316842> in <module>
----> 1 gen_window("bh7", 1)

~/Documents/school/penn/software/uvtools/uvtools/dspec.py in gen_window(window, N, alpha, edgecut_low, edgecut_hi, **kwargs)
    461         a_k = [0.27105140069342, -0.43329793923448, 0.21812299954311, -0.06592544638803,
    462                0.01081174209837, -0.00077658482522, 0.00001388721735]
--> 463         w[edgecut_low:edgecut_hi] = np.sum([[ak * np.cos(2 * np.pi * k * n / float(N - Ncut - 1)) for n in range(N - Ncut)] for k, ak in enumerate(a_k)], axis=0)
    464     else:
    465         try:

~/Documents/school/penn/software/uvtools/uvtools/dspec.py in <listcomp>(.0)
    461         a_k = [0.27105140069342, -0.43329793923448, 0.21812299954311, -0.06592544638803,
    462                0.01081174209837, -0.00077658482522, 0.00001388721735]
--> 463         w[edgecut_low:edgecut_hi] = np.sum([[ak * np.cos(2 * np.pi * k * n / float(N - Ncut - 1)) for n in range(N - Ncut)] for k, ak in enumerate(a_k)], axis=0)
    464     else:
    465         try:

~/Documents/school/penn/software/uvtools/uvtools/dspec.py in <listcomp>(.0)
    461         a_k = [0.27105140069342, -0.43329793923448, 0.21812299954311, -0.06592544638803,
    462                0.01081174209837, -0.00077658482522, 0.00001388721735]
--> 463         w[edgecut_low:edgecut_hi] = np.sum([[ak * np.cos(2 * np.pi * k * n / float(N - Ncut - 1)) for n in range(N - Ncut)] for k, ak in enumerate(a_k)], axis=0)
    464     else:
    465         try:

ZeroDivisionError: float division by zero
```

I think that requesting a window of size 1 is not a crazy thing, so I'd like to return an array with a single value of 1. The scipy functions behave this way (and will even return an empty array for N=0, though I don't necessarily think we need to support that), so I'd like to keep that parallelism so the user has more consistency.

nkern: Sounds good.

nkern:
@plaplant ok, found a nice solution to your issue which is to use `scipy.signal.windows.general_cosine` to generate the cosine family windows which automatically incorporates your desire to have an `N=1` use-case. lmk if you see anything else, thanks!

PR #35, opened 2019-06-12
Linear clean

anguta:
Implementation of linear clean for use in analysis of visibility data. The sinc_downweight method will eventually be referenced by hera_pspec (and the version therein deprecated). 

I've also added a (hopefully?) helpful tutorial notebook in "examples". 

cheers,

-Aaron


uvtools/dspec.py
```diff
@@ -775,3 +868,71 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False,
...
+        float or list of floats of widths of delay filter windows in nanosec
+    filter_factors: float or list
+        float or list of floats of filtering factors.
+    cache: dictionary, optional dictionary storing filter matrices with keys
+    (nchan, df, ) + (filter_centers) + (filter_widths) + \
+    (filter_factors)
+
+    !!!-------------
+    WARNING: The following parameters are intended for studying weighting
+    but shouldn't not be changed from defaults in practical data analysis!
```
nkern: typo: shouldn't not be

uvtools/dspec.py
```diff
@@ -775,3 +868,71 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False,
         mdl_array[i, :] = bf_model
 
     return mdl_array, cn_array, inp_data
+
+
+def sinc_downweight_mat_inv(nchan, df, filter_centers, filter_widths,
+                            filter_factors, cache = {}, wrap = False, wrap_interval=1,
+                            nwraps = 1000, no_regularization = False):
+    """
+    Computes inverse of clean weights for a baseline.
```
nkern: did we say we weren't referencing "clean", and is this a holdout of that? I forget

anguta: you're right. I'm removing the reference.

uvtools/dspec.py
anguta: Thanks for catching this. I've fixed it.

PR #40, opened 2019-07-09
fixed installation order issues

steven-murray:
Should fix the installation order issue in the same way that `pyuvdata` did recently.

Also adds some installation instructions and an `environment.yml` for simpler installation of dev environment.

### DIFF
--- a/uvtools/version.py
+++ b/uvtools/version.py
@@ -5,11 +5,13 @@
 
 from __future__ import absolute_import, division, print_function
 
-import os, sys
-import subprocess
 import json
+import os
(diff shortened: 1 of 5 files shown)


uvtools/version.py
```diff
@@ -5,10 +5,11 @@
 
 from __future__ import absolute_import, division, print_function
 
-import os
+import os, sys
```
bhazelton: these should be on separate lines per PEP8

steven-murray: Yup my bad. Fixed now.

PR #41, opened 2019-08-09
Tidy-up plotting functions and add docstrings

philbull:
This PR removes some obsolete/buggy lines of code, adds `vmin` and `vmax` kwargs to plotting functions (alongside the more cryptic `mx` and `drng` kwargs), adds in-line comments, and adds extensive docstrings to all plotting functions.

### DIFF
--- a/uvtools/plot.py
+++ b/uvtools/plot.py
@@ -28,7 +28,6 @@ def data_mode(data, mode='abs'):
         data = np.angle(data)
     elif mode.startswith('abs'):
         data = np.absolute(data)
-        #data = np.masked_less_equal(data, 0)
     elif mode.startswith('real'):
         data = data.real
     elif mode.startswith('imag'):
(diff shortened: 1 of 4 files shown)


uvtools/plot.py
```diff
@@ -2,115 +2,385 @@
...
+    
+    vmin, vmax : float, optional
+        Minimum and maximum values of the color scale. If not set (and `mx` and 
+        `drng` are not set), the min. and max. values of the data will be used.
+        
+        Note that that the min. and max. values are the ones _after_ the data 
+        have been transformed according to `mode`. So, if `mode='log'`, these 
+        values are the min. and max. of log_10(data).
+        
+    mx : float, optional (deprecated)
```
adampbeardsley: It seems odd that this and `drng` are said to be deprecated, but they are still working options. In fact, the code internally uses `mx` and `drng` even if `vmin` and `vmax` are provided.

uvtools/plot.py
```diff
@@ -2,115 +2,385 @@
...
 def plot_hmap_ortho(h, cmap='jet', mode='log', mx=None, drng=None, 
-        res=0.25, verbose=False, normalize=False):
+                    res=0.25, verbose=False, normalize=False):
+    """
+    Plot a Healpix map in ortho projection.
+    
+    Parameters
+    ----------
+    h : healpix map
+        Healpy HEALPIX map.
```
adampbeardsley: Could you specify further what is expected here? Is this a specific object from healpy? On line 201 it looks like it's some kind of interpolation function.

philbull: On closer inspection, it turns out to be an aipy `HealpixMap` instance! I've changed the string.

uvtools/plot.py
philbull: I've removed the "deprecated" language. This function is widely used in many packages, so I wanted to maintain backwards compatibility.

philbull:
@adampbeardsley Thanks! I've made the requested changes and also enabled a few more tests.

PR #43, opened 2019-09-05
Linear clean fringe rate

anguta:
Added fringe-rate filtering capabilities to linear_delay_clean. Due to covariance matrix inversion limitations, linear fringe-rate filter only supports separable cleans (a clean along time and frequency dimensions independently) rather then a full 2d clean.


anguta:
I've added hooks in vis_clean, fringe_clean etc... so that linear methods can be called from hera_cal FringeRateFilter.

uvtools/dspec.py
```diff
@@ -62,8 +62,8 @@ def calc_width(filter_size, real_delta, nsamples):
 
 
 def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False, tol=1e-9, window='none',
-                             skip_wgt=0.1, maxiter=100, gain=0.1, filt2d_mode='rect', alpha=0.5,
-                             edgecut_low=0, edgecut_hi=0, add_clean_residual=False):
+                             skip_wgt=0.1, maxiter=100, gain=0.1, filt2d_mode='rect', alpha=0.5, linear = False,
```
nkern: try using the same convention as every other kwarg here: `linear=False` i.e. no spaces between assign operators in a function signature :)

uvtools/dspec.py
```diff
@@ -216,12 +271,21 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False,
...
+            d_res = np.fft.fft2(_d_res)
+        else:
+            d_res = np.fft.fft(_d_res, axis = 0)
+
 
     return d_mdl, d_res, info
 
-def linear_delay_filter(data, wgts, df, filter_centers, filter_widths, filter_factors, cache = {}):
+def linear_delay_filter(data, wgts, df, filter_centers, filter_widths, filter_factors, cache = {},
+                        clean_dimensions = [False, True]):
```
nkern: I thought we were avoiding CLEAN terminology? perhaps `filter_dimensions`?

nkern:
Nitpicky, but please conform the code to PEP8 and the conventions adopted here, namely: no spaces between assign operator in a function signature i.e. `..., linear=True,` not `..., linear = True,`.

uvtools/dspec.py
```diff
@@ -62,8 +62,8 @@ def calc_width(filter_size, real_delta, nsamples):
 
 
 def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False, tol=1e-9, window='none',
-                             skip_wgt=0.1, maxiter=100, gain=0.1, filt2d_mode='rect', alpha=0.5,
-                             edgecut_low=0, edgecut_hi=0, add_clean_residual=False):
+                             skip_wgt=0.1, maxiter=100, gain=0.1, filt2d_mode='rect', alpha=0.5, linear = False,
+                             edgecut_low=0, edgecut_hi=0, add_clean_residual=False, cache = {}):
     '''Apply a highpass fourier filter to data. Uses aipy.deconv.clean. Default is a 1D clean
```
nkern: should update this to read uses aipy.deconv.clean if linear == False. also perhaps better to put `linear` kwarg next to `cache` in the function signature, and same fro docstring

uvtools/dspec.py
```diff
@@ -150,9 +155,24 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False,
...
-            _d_cl, info = aipy.deconv.clean(_d, _w, area=area, tol=tol, stop_if_div=False, maxiter=maxiter, gain=gain)
-            _d_res = info['res']
-            del info['res']
+            if not linear:
+                _d_cl, info = aipy.deconv.clean(_d, _w, area=area, tol=tol, stop_if_div=False, maxiter=maxiter, gain=gain)
+                _d_res = info['res']
+                del info['res']
+            else:
+                ff = [ tol ]
+                if isinstance(filtersize, np.float) == 1:
```
nkern: this should just be `if isinstance(filtersize, float):`

uvtools/dspec.py
```diff
@@ -150,9 +155,24 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False,
...
+                    fc = [ 0. ]
+                    fw = [ filter_size ]
+                else:
+                    total_width = filter_size[1]-filter_size[0]
```
nkern: does the `calc_width` function (used above) do the same thing for you? If so consider using that, otherwise no problem.

uvtools/dspec.py
```diff
@@ -294,21 +380,53 @@ def linear_delay_filter(data, wgts, df, filter_centers, filter_widths, filter_fa
...
+            tuple(filter_widths[1]) + tuple(filter_factors[1]) + tuple(wght.tolist()) + ('inverse',)
+            if not filter_key in cache:
+                try:
+                    cache[filter_key] = np.linalg.pinv(filter_mat)
+                except LinearAlgError:
+                    #if SVD fails to converge, set filter matrix to zeros.
+                    cache[filter_key] = np.zeros((nf,nf), dtype = complex)
+                    skipped_1.append(sample_num)
+            filter_mat = cache[filter_key]
+            #print(np.std(sample.real))
```
nkern: eliminate debugging lines

uvtools/dspec.py
```diff
@@ -294,21 +380,53 @@ def linear_delay_filter(data, wgts, df, filter_centers, filter_widths, filter_fa
...
-        output[sample_num] = np.dot(filter_mat, sample)
-        #print(np.std(output[sample_num].real))
-    if data_1d:
-        output = output[0]
-    return output
+    ntimes = data.shape[0]
+    info = {'filter_centers':filter_centers, 'filter_widths':filter_widths, 'filter_factors': filter_factors}
+    skipped_0 = []
+    skipped_1 = []
+    if clean_dimensions[1]:
```
nkern: it would be helpful to add at least a few comments over the next dozen or so lines to describe what you are doing, in case someone else needs to update this code

nkern:
Looks pretty good to me, just some minor comments. lmk when you have them done and I'll have on e last review, thanks!

uvtools/dspec.py
anguta: Good point. I'll go ahead and change this.

anguta:
> Nitpicky, but please conform the code to PEP8 and the conventions adopted here, namely: no spaces between assign operator in a function signature i.e. `..., linear=True,` not `..., linear = True,`.

Yep. I've gone through and removed spaces at the equal signs in function arguments.

uvtools/dspec.py
anguta: I've rearranged the arg order to reflect the doc string.

uvtools/dspec.py
anguta: yeah. I don't know how that happened... Changed.

anguta:
> Nitpicky, but please conform the code to PEP8 and the conventions adopted here, namely: no spaces between assign operator in a function signature i.e. `..., linear=True,` not `..., linear = True,`.

No worries, I've fixed these.

uvtools/dspec.py
anguta: no more print statements :)

uvtools/dspec.py
anguta: I've added some comments here and cleaned up the code :)

anguta:
> Nitpicky, but please conform the code to PEP8 and the conventions adopted here, namely: no spaces between assign operator in a function signature i.e. `..., linear=True,` not `..., linear = True,`.

I've fixed all of the pep8 equals violations I could find.

anguta:
I've added some more functionality to add "deconvolved" foregrounds by applying clean or linear-least-squares clean to the difference between the original data and the residual of the linear filter. The foreground restoration with linear-leastsq clean gives a fully linear estimator for foregrounds/signal that removes flagging sidelobes.

uvtools/dspec.py
```diff
@@ -104,6 +105,27 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False,
         add_clean_residual : bool, if True, adds the CLEAN residual within the CLEAN bounds
             in fourier space to the CLEAN model. Note that the residual actually returned is
             not the CLEAN residual, but the residual in input data space.
+        linear : bool,
+                 use aipy.deconv.clean if linear == False
+                 if True, perform linear delay filtering.
+        cache : dict, optional dictionary for storing pre-computed filtering matrices in linear
+            cleaning.
+        deconv_linear_foregrounds : bool, if True, then apply clean to data - residual where
+                                          res is the data-vector after applying a linear clean filter.
```
nkern: use a single indent beyond the kwarg for line continuations. same for kwargs below.

uvtools/dspec.py
```diff
@@ -189,21 +275,87 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False,
...
 
+        uthresh_fg, lthresh_fg = calc_width(fg_restore_size[0], real_delta[0], data.shape[0])
+        a1fg = np.ones(data.shape[0], dtype=np.int)
+        a1fg[uthresh_fg:lthresh_fg] = 0
+        uthresh_fg, lthresh_fg = calc_width(fg_restore_size[1], real_delta[1], data.shape[1])
+        a2fg = np.ones(data.shape[-1], dtype=np.int)
+        a2fg[uthresh_fg:lthresh_fg] = 0
+        area_fg = np.outer(a1fg, a2fg)
+
+        fc = []
```
nkern: please add a comment here that states that this block is only needed for `linear filter`

uvtools/dspec.py
```diff
@@ -189,21 +275,87 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False,
...
+                    nmin = int((fcfg[1][0] - fwfg[1][0]) * real_delta[1] * data.shape[-1])
+                    nmax = int((fcfg[1][0] + fwfg[1][0]) * real_delta[1] * data.shape[-1])
+                    info['fg_deconv'] = {'method':'leastsq', 'nmin':nmin, 'nmax':nmax}
+                    d_cl, _, _ = delay_filter_leastsq(data * wgts * win - d_r, flags=wgts==0., sigma=1.,
+                                                      nmax=(nmin, nmax), freq_units=True, even_modes=True)
+                    _d_cl = np.fft.ifft(d_cl)
+
+            else:
+                _d_cl = _d - _d_res
+
```
nkern: unnecessary line breaks below

uvtools/dspec.py
```diff
@@ -212,63 +364,97 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False,
...
     if clean2d:
         d_mdl = np.fft.fft2(_d_cl, axes=(0, 1))
+        d_res = np.fft.fft2(_d_res, axes=(0, 1))
     else:
         d_mdl = np.fft.fft(_d_cl)
+        d_res = np.fft.fft(_d_res)
 
-    # get residual in data space
-    d_res = (data - d_mdl) * ~np.isclose(wgts * win, 0.0)
+    if not linear:
```
nkern: why get rid of the comment?

uvtools/dspec.py
```diff
@@ -212,63 +364,97 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False,
...
+        d_res = (data - d_mdl) * ~np.isclose(wgts * win, 0.0)
 
     return d_mdl, d_res, info
 
-def linear_delay_filter(data, wgts, df, filter_centers, filter_widths, filter_factors, cache = {}):
+def linear_filter(data, wgts, delta_data, filter_dimensions, filter_centers, filter_widths, filter_factors, cache = {}):
     '''Apply a linear delay filter to waterfall data.
+        Due to performance reasons, linear cleaning only supports separable delay-rate filters.
```
nkern: I think we now use the `delay / fringe-rate` terminology instead of `delay-rate`

uvtools/dspec.py
```diff
@@ -212,63 +364,97 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False,
...
-        filter_widths = list(filter_widths)
-    if isinstance(filter_factors, np.ndarray):
-        filter_factors = list(filter_factors)
-    if isinstance(filter_centers, np.float):
-        filter_centers = [filter_centers]
-    if isinstance(filter_widths, np.float):
-        filter_widths = [filter_widths]
-    if isinstance(filter_factors, np.float):
-        filter_factors = [filter_factors]
+    assert isinstance(filter_dimensions, (list,tuple,int)), "filter_dimensions must be a list or tuple"
```
nkern: the next 50 lines are quite hard to parse, please add comments for future people who might need to maintain this code. also some PEP8 non-conformity here: spaces between operator, after commas..

uvtools/dspec.py
```diff
@@ -292,32 +478,67 @@ def linear_delay_filter(data, wgts, df, filter_centers, filter_widths, filter_fa
...
-        #print(np.std(sample.real))
-        #print(np.diag(filter_mat))
-        output[sample_num] = np.dot(filter_mat, sample)
-        #print(np.std(output[sample_num].real))
-    if data_1d:
+    ntimes = data.shape[0]
+    info = {'filter_centers':filter_centers, 'filter_widths':filter_widths, 'filter_factors': filter_factors,
+            'delta_data':delta_data, 'data_shape':data.shape, 'filter_dimensions': filter_dimensions}
+    skipped = [[],[]]
+    #in the lines below, we iterate over the time dimension. For each time, we
```
nkern: space after `#`

nkern:
@anguta a few requests to `clean` up the code no pun intended :), including PEP8, doc-strings and comments. 

Also seems like you don't have any tests for the new fg_restore functionality, but please correct me if I am wrong.

uvtools/dspec.py
anguta: I've fixed instances with too many indents.

uvtools/dspec.py
anguta: there is now one line break here.

uvtools/dspec.py
anguta: I've changed this.

uvtools/dspec.py
anguta: I've eliminated instances with no space after `#`

uvtools/dspec.py
anguta: I've cleaned up domentation here and added pep8 conformity.

anguta:
> @anguta a few requests to `clean` up the code: pun intended :), including PEP8, doc-strings and comments.
> 
> Also seems like you don't have any tests for the new fg_restore functionality, but please correct me if I am wrong.

There are numerous tests checking the fg_restore functionality in 

`test_vis_filter_linear`

nkern:
@anguta thanks for cleaning up the code! This looks good to me so long as you can get travis to pass: it looks like the failures are not something trivial like IERS, so once you have that resolved I'm good to merge. As a general question about the FG restore capability: do you have any concerns about applying functionally different filters for the low and high delay modes? Is there possibility of losing some signal, say, if the FG restore window doesn't exactly line up flush with the linear sinc downweight window? Are there also possibilities of the FG CLEAN to insert residual errors into the high delay modes as well? Good to get your thoughts here on this for posterity, thanks!

uvtools/dspec.py
anguta: Done.

uvtools/dspec.py
anguta: This was a mistake. I've readded the comment.

anguta:
> @anguta thanks for cleaning up the code! This looks good to me so long as you can get travis to pass: it looks like the failures are not something trivial like IERS, so once you have that resolved I'm good to merge. As a general question about the FG restore capability: do you have any concerns about applying functionally different filters for the low and high delay modes? Is there possibility of losing some signal, say, if the FG restore window doesn't exactly line up flush with the linear sinc downweight window? Are there also possibilities of the FG CLEAN to insert residual errors into the high delay modes as well? Good to get your thoughts here on this for posterity, thanks!

There will be missing foreground power from two different effects. The first, which is present in Iterative clean, arises from the fact that limiting the reconstructed Fourier modes to be harmonics of the bandwidth, leads to intrinsic information loss. Secondly, if the filtering region is larger then the reconstruction region, these delays will have signal loss. Since this loss is linear, one should be able to account for it in the 21cm signal (which is described by bandpowers) during the normalization power-spectrum step. The loss in power that is not described by bandpowers (such as the foregrounds) will be irreversible though. I'm not sure what is meant by FG CLEAN but as long as the filtering matrix is well conditioned, it should not introduce high delay errors. There may be a perverse case that I haven't encountered of course so we should always be diligent with validation.

PR #46, opened 2019-11-14
Diff plotter

r-pascua:
This PR implements two new plotting tools for visualizing differences between a pair of visibilities: `plot_diff_waterfall` and `plot_diff_uv`. Both functions act on a pair of `UVData` objects to produce a set of plots visualizing the differences of their data arrays (or a subset thereof). `plot_diff_waterfall` exclusively plots differences for a particular baseline and polarization, and has the capability of plotting the differences in four different spaces (time-frequency, time-delay, fringe rate-frequency, fringe rate-delay), either simultaneously or some combination of the four. `plot_diff_uv` plots the difference for the entire data array, visualizing the results in the uv-plane. Both functions visualize the differences by plotting the amplitude differences alongside the phase differences (in different plots on the same figure). In addition to these two functions, a handful of helper functions were developed in the `utils` module.


r-pascua:
Important note: as of yet, no tests have been developed for the new tools implemented here, as it's unclear how to automatically test a plotting suite for correctness.

uvtools/plot.py
```diff
@@ -466,3 +469,265 @@ def omni_view_gif(filenames, name='omni_movie.gif'):
...
+
+            # plot waterfall and add a colorbar
+            fig.sca(ax)
+            cax = waterfall(diff, mode="real", cmap='viridis', 
+                            extent=extent(xvals, yvals))
+            fig.colorbar(cax)
+
+    # display the figure
+    plt.tight_layout()
+    plt.show()
```
philbull: Best to let the user do this themselves, as otherwise it will make it harder to use this function in an automated way. In a Jupyter notebook, returning a `Figure` should result in the figure being plotted automatically anyway, so probably best to do that.

uvtools/plot.py
```diff
@@ -466,3 +469,265 @@ def omni_view_gif(filenames, name='omni_movie.gif'):
...
+        Which spaces to use for investigating differences. Available options 
+        are as follows: time and frequency ('time_vs_freq'); time and delay 
+        ('time_vs_dly'); fringe rate and frequency ('fr_vs_freq'); fringe 
+        rate and delay ('fr_vs_dly'). Default is to use all plot types.
+    
+    skip_check : bool, optional
+        Whether to check that the metadata in `uvd1` and `uvd2` match.
+        Default behavior is to check the metadata.
+
+    save_path : str, optional
```
philbull: It's probably better to return the `matplotlib.Figure` object and allow the user to save it. It's only one extra line of code for the user, and gives them a lot more control over the figure.

uvtools/plot.py
```diff
@@ -466,3 +469,265 @@ def omni_view_gif(filenames, name='omni_movie.gif'):
...
+        visibility waterfalls. See pyuvdata.UVData.get_data method docstring 
+        for information on accepted tuples.
+    
+    plot_type : str, tuple of str, or list of str, optional
+        Which spaces to use for investigating differences. Available options 
+        are as follows: time and frequency ('time_vs_freq'); time and delay 
+        ('time_vs_dly'); fringe rate and frequency ('fr_vs_freq'); fringe 
+        rate and delay ('fr_vs_dly'). Default is to use all plot types.
+    
+    skip_check : bool, optional
```
philbull: This could be given a more obvious name. `check_metadata` perhaps? It's also a little unclear about what happens if they don't match (e.g. one has more time samples than the other) -- is an error raised? What happens if the check is turned off, but they don't match?

uvtools/plot.py
```diff
@@ -466,3 +469,265 @@ def omni_view_gif(filenames, name='omni_movie.gif'):
...
+def plot_diff_waterfall(uvd1, uvd2, antpairpol, plot_type="all", 
+                        skip_check=False, save_path=None):
+    """Produce waterfall plot(s) of differenced visibilities.
+
+    Parameters
+    ----------
+    uvd1, uvd2 : pyuvdata.UVData
+        UVData objects which store visibilities to be differenced and their
+        associated metadata. They should have the same number of frequencies,
+        same baselines, and same times as each other.
```
philbull: If you want to be fancy, it's possible to use the `UVData.select()` method to obtain matching shapes for the two objects. You just need to call `numpy.intersect1d` on the various metadata arrays to see which elements are common to both UVData objects.

uvtools/plot.py
```diff
@@ -466,3 +469,265 @@ def omni_view_gif(filenames, name='omni_movie.gif'):
...
+    plt.show()
+
+    # save if desired
+    if save_path is not None:
+        utils.savefig(fig, save_path)
+
+def plot_diff_uv(uvd1, uvd2, pol=None, speedup=True,
+                 skip_check=False, save_path=None,
+                 resolution=50):
+    """Summary plot for difference between visibilities.
```
philbull: This could be a bit more descriptive. For example, you could mention that it's plotted in the uv plane, and what the quantity being plotted in each bin is (i.e. how is the difference being averaged etc.).
Also, it's definitely worth mentioning that it's slow (you can give a typical run time for a typical number of blpairs).

uvtools/plot.py
```diff
@@ -466,3 +469,265 @@ def omni_view_gif(filenames, name='omni_movie.gif'):
...
+
+    skip_check : bool, optional
+        Whether to check that the metadata for `uvd1` and `uvd2` match.
+        Default behavior is to check the metadata.
+
+    save_path : str, optional
+        Path to where the figure should be saved; may be absolute or relative.
+        Default is to not save the figure.
+    
+    resolution : int, optional
```
philbull: I would call this `bins` instead of `resolution`, to match better with the usage of the matplotlib histogram functions.

uvtools/plot.py
```diff
@@ -466,3 +469,265 @@ def omni_view_gif(filenames, name='omni_movie.gif'):
...
+        slower implementation. The faster implementation does not produce
+        figures as nice as the slower implementation, but the slower
+        implementation is only really usable for small sets of data.
+        Default is to use the fast implementation.
+
+    skip_check : bool, optional
+        Whether to check that the metadata for `uvd1` and `uvd2` match.
+        Default behavior is to check the metadata.
+
+    save_path : str, optional
```
philbull: Same as above; better to let the user do this themselves.

uvtools/plot.py
```diff
@@ -466,3 +469,265 @@ def omni_view_gif(filenames, name='omni_movie.gif'):
...
+        `uvd1` and `uvd2`. Default is to use the 0th polarization.
+
+    speedup : bool, optional
+        Whether to use the fast implementation of this plotting tool or the
+        slower implementation. The faster implementation does not produce
+        figures as nice as the slower implementation, but the slower
+        implementation is only really usable for small sets of data.
+        Default is to use the fast implementation.
+
+    skip_check : bool, optional
```
philbull: Same as above.

uvtools/plot.py
```diff
@@ -466,3 +469,265 @@ def omni_view_gif(filenames, name='omni_movie.gif'):
...
+    uvd1, uvd2 : pyuvdata.UVData
+        Input UVData objects which contain the visibilities to be differenced 
+        and any relevant metadata.
+
+    pol : str, None, optional
+        String specifying which polarization to be used. Must be one of the
+        polarizations listed in the UVData.get_pols() method for both
+        `uvd1` and `uvd2`. Default is to use the 0th polarization.
+
+    speedup : bool, optional
```
philbull: Might be worth explaining what the fast method is doing that makes it fast, so users can understand if this might cause issues in how they interpret the plots.

uvtools/plot.py
```diff
@@ -466,3 +469,265 @@ def omni_view_gif(filenames, name='omni_movie.gif'):
...
+
+        # plot stuff
+        for ax, diff in zip(axes[i], diffs):
+            # set labels
+            ax.set_xlabel(xlabel, fontsize=12)
+            ax.set_ylabel(ylabel, fontsize=12)
+
+            # plot waterfall and add a colorbar
+            fig.sca(ax)
+            cax = waterfall(diff, mode="real", cmap='viridis',
```
philbull: Better to use a cyclic colormap for the phases (e.g. `twilight`), as otherwise differences close to a phase wrap look large when they are actually small.

uvtools/utils.py
```diff
@@ -99,3 +100,128 @@ def search_data(templates, pols, matched_pols=False, reverse_nesting=False, flat
...
+    Returns
+    -------
+    data_fft : np.ndarray
+        The Fourier transform of the data along the specified axis. The array
+        has the same shape as the original data, with the same ordering.
+    """
+
+    return fftshift(fft(fftshift(data, axis), axis=axis), axis)
+
+def get_fourier_freqs(times):
```
philbull: The `get` prefix is normally used for methods in classes. `fourier_freqs` is fine here.

uvtools/utils.py
```diff
@@ -99,3 +100,128 @@ def search_data(templates, pols, matched_pols=False, reverse_nesting=False, flat
...
+    """
+    if mode == "abs":
+        return np.abs(vis1) - np.abs(vis2)
+    elif mode == "phs":
+        return np.angle(vis1) - np.angle(vis2)
+    elif mode == "complex":
+        return np.abs(vis1 - vis2)
+    else:
+        raise ValueError("You have not specified an accepted differencing "
+                         "mode. Please see the documentation for the "
```
philbull: May as well just say which modes are acceptable here. There's only 3.

uvtools/plot.py
```diff
@@ -466,3 +469,265 @@ def omni_view_gif(filenames, name='omni_movie.gif'):
...
+
+    # import astropy constants to convert freq to wavelength
+    from astropy.constants import c
+    wavelengths = c.value / freqs
+
+    # get uvw vectors; shape = (Nfreq, Nblts, 3)
+    uvw_vecs = np.array([bl_vecs / wavelength for wavelength in wavelengths])
+    
+    # reshape uvw vectors to (Nblts, Nfreq, 3)
+    uvw_vecs = np.einsum("ijk->jik", uvw_vecs)
```
philbull: `einsum` is occasionally stupid and does operations in a brute force way. Might be worth checking if `numpy.swapaxes` is faster in this particular case.

philbull:
> Important note: as of yet, no tests have been developed for the new tools implemented here, as it's unclear how to automatically test a plotting suite for correctness.

Take a look at the plotting tests in `hera_pspec` (https://github.com/HERA-Team/hera_pspec/blob/master/hera_pspec/tests/test_plot.py). This is a reasonably advanced way of doing things -- inspecting the plots to make sure the right number and type of elements is present. It would be fine to do something simpler here, like just calling the plotting functions and checking thagt a `Figure` is returned. You can also check to make sure error conditions actually raise the right errors.

uvtools/plot.py
r-pascua: The phase plot is actually a plot of the difference of phases, not the phase of the complex difference. So I don't think a cyclic colormap is appropriate here, especially given that the `vmin` and `vmax` values are not hard-coded to be -2pi and 2pi. Now, if we were plotting the phase of the complex difference, then I would make sure to use a cyclic colormap and fix the `vmin` and `vmax` values to -pi and pi. (I can add a column for the phase of the complex difference if you think it would be useful to also have that.)

uvtools/plot.py
r-pascua: Ah, this used to actually do something (when I had two implementations for plotting), but now there's only one way of generating the arrays to plot. The new method is also significantly faster than what the faster method of the two old implementations were. (It now uses `scipy.stats.binned_statistic_2d`. The previous implementation wasn't exactly doing the right thing.)

uvtools/plot.py
r-pascua: I updated the docstring to explain what is actually being plotted. Let me know if the new docstring is descriptive enough, or if I need to add more detail.

uvtools/plot.py
r-pascua: Hm. It looks like `numpy.intersect1d` is pretty conservative when making comparisons, only saying things intersect if the values are equal to one part in ~10^16. I suspect that if I were to use `numpy.intersect1d` in conjunction with `UVData.select()` to make things have the same shape, then it would often be the case that this ends up not plotting anything at all. In fact, the `check_metadata` conditions are much more lenient than this, just looking to see that the time arrays agree to a mean integration time and the frequency arrays agree to a mean channel width. (The check on the baseline vectors is a bit more stringent, using the default tolerance of `numpy.isclose`.)

philbull: OK, sounds fine to me in that case.

uvtools/plot.py
philbull: Looks good!

PR #50, opened 2019-12-03
Restoration matrix

anguta:
Calculate a matrix that performs linear foreground in-painting (similar to the linear_lsq clean method but provides the linear operator matrix that can be propagated into statistics calculations).

### DIFF
--- a/uvtools/dspec.py
+++ b/uvtools/dspec.py
@@ -837,7 +837,7 @@ def _get_bl_dly(bl_len, horizon=1., standoff=0., min_dly=0.):
     return bl_dly
 
 
-def gen_window(window, N, alpha=0.5, edgecut_low=0, edgecut_hi=0, normalization='none'', **kwargs):
+def gen_window(window, N, alpha=0.5, edgecut_low=0, edgecut_hi=0, normalization='none', **kwargs):
     """
     Generate a 1D window function of length N.
 
--- a/uvtools/dspec.py
+++ b/uvtools/dspec.py
@@ -551,7 +551,7 @@ def linear_filter(data, wgts, delta_data, filter_dimensions, filter_centers, fil
             else:
                 skipped[fs-1].append(sample_num)
 
-    #1d data will only be filtered across "channels". 
+    #1d data will only be filtered across "channels".
     if data_1d and ntimes == 1:
         output = output[0]
     info['skipped_time_steps'] = skipped[0]
@@ -1247,7 +1247,7 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, freq_units =
 
 def delay_interpolation_matrix(nchan, ndelay, wgts, fundamental_period=None, cache={}, taper='none', return_diagnostics=False):
     """
-    Copute a foreground interpolation matrix.
+    Compute a foreground interpolation matrix.
 
     Computes a foreground interpolation matrix that, when applied to data,
     interpolates over flagged channels with delays between
@@ -1268,7 +1268,7 @@ def delay_interpolation_matrix(nchan, ndelay, wgts, fundamental_period=None, cac
     ----------
     nchan: int
         Number of frequency channels to interpolate over.
-    ndelay,
+    ndelay: int
         number of delays to use in interpolation.
     wgts: float array
         wgts to be applied to each frequency channel.
--- a/uvtools/dspec.py
+++ b/uvtools/dspec.py
@@ -837,7 +837,7 @@ def _get_bl_dly(bl_len, horizon=1., standoff=0., min_dly=0.):
     return bl_dly
 
 
-def gen_window(window, N, alpha=0.5, edgecut_low=0, edgecut_hi=0, normalize='none'', **kwargs):
+def gen_window(window, N, alpha=0.5, edgecut_low=0, edgecut_hi=0, normalization='none'', **kwargs):
     """
     Generate a 1D window function of length N.
 
@@ -893,9 +893,9 @@ def gen_window(window, N, alpha=0.5, edgecut_low=0, edgecut_hi=0, normalize='non
             w[edgecut_low:edgecut_hi] = getattr(windows, window)(N - Ncut)
         except AttributeError:
             raise ValueError("Didn't recognize window {}".format(window))
-    if normalize == 'rms':
+    if normalization == 'rms':
         w = w / np.sqrt(np.mean(np.abs(w)**2.))
-    if normalize == 'mean':
+    if normalization == 'mean':
         w = w / w.mean()
     return w
 
--- a/uvtools/dspec.py
+++ b/uvtools/dspec.py
@@ -196,7 +196,7 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False,
                 del info['res']
             else:
                 d_r, info = linear_filter(data * wgts * win, wgts * win, delta_data=real_delta,
-                                                filter_dimensions = [1], filter_centers=fc, filter_widths=fw, filter_factors=ff, cache=cache)
+                                                filter_dimensions = [1], filter_centers=fc, filter_half_widths=fw, filter_factors=ff, cache=cache)
                 _d_res = np.fft.ifft(d_r)
                 if deconv_linear_foregrounds:
                     if fg_deconv_method == 'clean':
@@ -234,7 +234,7 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False,
                         info.append(info_here)
                     else:
                         d_r, info_here = linear_filter(data[i] * wgts[i] * win, wgts[i] * win, delta_data=real_delta,
-                                                            filter_dimensions=[1], filter_centers=fc, filter_widths=fw, filter_factors=ff, cache=cache)
+                                                            filter_dimensions=[1], filter_centers=fc, filter_half_widths=fw, filter_factors=ff, cache=cache)
                         _d_res[i] = np.fft.ifft(d_r)
                         if deconv_linear_foregrounds:
                             if fg_deconv_method == 'clean':
@@ -331,7 +331,7 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False,
         else:
             assert filt2d_mode == "plus", "2d linear deconvolution only supports filt2d_mode == 'plus'."
 
-            d_r, info = linear_filter(data * wgts * win, wgts * win, delta_data=[real_delta[0],real_delta[1]], filter_centers=fc, filter_widths=fw,
+            d_r, info = linear_filter(data * wgts * win, wgts * win, delta_data=[real_delta[0],real_delta[1]], filter_centers=fc, filter_half_widths=fw,
                                          filter_factors=ff, cache=cache, filter_dimensions=[0, 1])
             _d_res = np.fft.ifft2(d_r)
             if deconv_linear_foregrounds:
@@ -369,7 +369,7 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False,
 
     return d_mdl, d_res, info
 
-def linear_filter(data, wgts, delta_data, filter_dimensions, filter_centers, filter_widths, filter_factors, cache = {}):
+def linear_filter(data, wgts, delta_data, filter_dimensions, filter_centers, filter_half_widths, filter_factors, cache = {}):
     '''Apply a linear delay filter to waterfall data.
         Due to performance reasons, linear filtering only supports separable delay/fringe-rate filters.
     Arguments:
@@ -382,12 +382,12 @@ def linear_filter(data, wgts, delta_data, filter_dimensions, filter_centers, fil
             list of integers indicating data dimensions to filter. Must be 0, 1, or -1
         filter_centers: float, list, or 1d numpy array of delays at which to center filter windows
             Typically in units of (seconds)
-        filter_widths: float, list, or 1d numpy array of widths of each delay filtere window
+        filter_half_widths: float, list, or 1d numpy array of widths of each delay filtere window
             with centers specified by filter_centers.
             Typically in units of (seconds)
         filter_factors: float, list, or 1d numpy array of factors by which filtering should be
             applied within each filter window specified in filter_centers and
-            filter_widths. If a float or length-1 list/ndarray is provided,
+            filter_half_widths. If a float or length-1 list/ndarray is provided,
             the same filter factor will be used in every filter window.
         cache: optional dictionary for storing pre-computed delay filter matrices.
 
@@ -444,15 +444,15 @@ def linear_filter(data, wgts, delta_data, filter_dimensions, filter_centers, fil
     # will only filter each dim a single time.
     # now check validity of other inputs. We perform the same check over multiple
     # inputs by iterating over a list with their names.
-    check_vars = [filter_centers, filter_widths, filter_factors]
-    check_names = ['filter_centers', 'filter_widths', 'filter_factors']
+    check_vars = [filter_centers, filter_half_widths, filter_factors]
+    check_names = ['filter_centers', 'filter_half_widths', 'filter_factors']
     for anum,avar in enumerate(check_vars):
         # If any of these inputs is a float or numpy array, convert to a list.
         if isinstance(avar, np.ndarray):
             check_vars[anum] = list(avar)
         elif isinstance(avar, np.float):
             check_vars[anum] = [avar]
-    filter_centers,filter_widths,filter_factors = check_vars
+    filter_centers,filter_half_widths,filter_factors = check_vars
     # Next, perform some checks that depend on the filtering dimensions provided.
     if 0 in filter_dimensions and 1 in filter_dimensions:
         for avar,aname in zip(check_vars,check_names):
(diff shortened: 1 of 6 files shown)


uvtools/dspec.py
```diff
@@ -837,7 +837,7 @@ def _get_bl_dly(bl_len, horizon=1., standoff=0., min_dly=0.):
     return bl_dly
 
 
-def gen_window(window, N, alpha=0.5, edgecut_low=0, edgecut_hi=0, **kwargs):
+def gen_window(window, N, alpha=0.5, edgecut_low=0, edgecut_hi=0, normalization='none', **kwargs):
```
plaplant: I think it's a little cleaner to set the default argument to normalization be the python keyword `None` rather than a string, but this also works.

uvtools/dspec.py
```diff
@@ -849,7 +849,10 @@ def gen_window(window, N, alpha=0.5, edgecut_low=0, edgecut_hi=0, **kwargs):
         edgecut_hi : int, number of bins to consider as zero-padded at the high-side
             of the array, such that the window smoothly connects to zero.
         alpha : if window is 'tukey', this is its alpha parameter.
+        normalize : str, optional
+            set to 'rms' to divide by rms and 'mean' to divide by mean.
     """
+    assert normalization in ['none', 'mean', 'rms'], "normalization must be one of ['none', 'rms', 'mean']"
```
plaplant: If you set the default argument to be `None` above, you can change this to:
```
if normalization is not None:
    assert normalization in ["mean", "rms"], <blah blah blah>
```
This makes it a little clearer what the allowed "actual values" are.

uvtools/dspec.py
```diff
@@ -890,7 +893,10 @@ def gen_window(window, N, alpha=0.5, edgecut_low=0, edgecut_hi=0, **kwargs):
             w[edgecut_low:edgecut_hi] = getattr(windows, window)(N - Ncut)
         except AttributeError:
             raise ValueError("Didn't recognize window {}".format(window))
-
+    if normalization == 'rms':
+        w = w / np.sqrt(np.mean(np.abs(w)**2.))
```
plaplant: As a personal style preference, I'd write this as `w /= np.sqrt(np.mean(np.abs(w)**2.))` (and similarly below), but this is fine too.

uvtools/dspec.py
```diff
@@ -1235,9 +1241,62 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, freq_units =
...
+def delay_interpolation_matrix(nchan, ndelay, wgts, fundamental_period=None, cache={}, taper='none', return_diagnostics=False):
+    '''
+    Computes a foreground interpolation matrix that, when applied to data,
+    interpolates over flagged channels with delays between -ndelays//2 and ndelays//2.
+
+    Parameters
+    ----------
+    nchan: integer
+        Number of frequency channels to interpolate over
+    ndelay,
```
plaplant: This should have a type (I'm assuming `int`?)

uvtools/dspec.py
plaplant: I'm also a big proponent of foregoing `assert` statements when sanitizing user input (because you can run python in such a way that they are ignored) and instead explicitly doing something like:
```
if normalization not in ["mean", "rms"]:
    raise AssertionError("My message goes here")
```
This makes sure potentially invalid user input is always caught.

uvtools/dspec.py
```diff
@@ -1235,9 +1241,62 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, freq_units =
...
+    cache: dict, optional
+        optional cache holding pre-computed matrices
+    taper: string, optional
+        use a taper to fit.
+    Returns
+    ----------
+    (nchan, nchan) numpy array
+        that can be used to interpolate over channel gaps.
+    '''
+    assert len(wgts) == nchan, "nchan must equal length of wgts"
```
plaplant: Again, if we're making sure the user passed us something valid, we should not be doing it with an `assert` statement.

uvtools/dspec.py
```diff
@@ -1235,9 +1241,62 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, freq_units =
...
+    matkey = (nchan, ndelay) + tuple(wgts)
+    assert np.sum((np.abs(wgts) > 0.).astype(float)) >= ndelay, "number of unflagged channels must be greater then or equal to number of delays"
```
plaplant: Same as above.

uvtools/dspec.py
```diff
@@ -1235,9 +1241,62 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, freq_units =
...
+    if fundamental_period is None:
+        fundamental_period = nchan #this tends to give well conditioned matrices.
```
plaplant: It might be worth putting this explanatory comment into the docstring as well.

uvtools/dspec.py
```diff
@@ -1235,9 +1241,62 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, freq_units =
...
+    ----------
+    (nchan, nchan) numpy array
+        that can be used to interpolate over channel gaps.
+    '''
+    assert len(wgts) == nchan, "nchan must equal length of wgts"
+    matkey = (nchan, ndelay) + tuple(wgts)
+    assert np.sum((np.abs(wgts) > 0.).astype(float)) >= ndelay, "number of unflagged channels must be greater then or equal to number of delays"
+    if fundamental_period is None:
+        fundamental_period = nchan #this tends to give well conditioned matrices. 
+    if not matkey in cache or return_diagnostics:
```
plaplant: I think this should be `if matkey not in cache...`

uvtools/dspec.py
```diff
@@ -1235,9 +1241,62 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, freq_units =
...
+        f, d = np.meshgrid(np.arange(nchan)-nchan/2, np.arange(-ndelay,ndelay), indexing='ij')
```
plaplant: I'm not a big fan of single-letter variable names. It makes it hard to search for them when trying to debug things.

uvtools/dspec.py
```diff
@@ -1235,9 +1241,62 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, freq_units =
...
+        #f = np.asarray([[chan for m in range(ndelay)] for chan in np.arange(-nchan//2,nchan//2)])
```
plaplant: Commented out code should be removed.

uvtools/dspec.py
```diff
@@ -1235,9 +1241,62 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, freq_units =
...
+        #d = np.asarray([np.arange(-ndelay//2, ndelay//2) for m in range(nchan)])
+        d = d / fundamental_period
+        a_mat = np.exp(2j * d * f * np.pi) / fundamental_period
+        wmat = np.diag(wgts * gen_window(taper, nchan)).astype(complex)
+        cmat=np.dot(a_mat.T, (a_mat.T * wgts).T)
+        if np.linalg.cond(cmat)>=1e9:
+            print('Warning!!!!: Poorly conditioned matrix! Your linear inpainting IS WRONG!'
```
plaplant: Rather than a print statement, this should be raised using the `warnings` module. (E.g., `warnings.warn("Poorly conditioned matrix! ...`)

uvtools/dspec.py
```diff
@@ -1235,9 +1241,62 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, freq_units =
...
+                  'Fix this by adjusting fundamental tones!')
+        cmati = np.linalg.inv(cmat)
+        tmat = np.dot(cmati,(a_mat.T * wgts))
+        #a_mat[:,0]=0.#for some reason edge coefficients are messed up
```
plaplant: Remove commented out code. Also, is it still a potential issue that the edge coefficients might be wrong?

uvtools/dspec.py
```diff
@@ -1250,12 +1309,12 @@ def sinc_downweight_mat_inv(nchan, df, filter_centers, filter_widths,
         channel width (Hz)
     filter_centers: float or list
         float or list of floats of centers of delay filter windows in nanosec
-    filter_widths: float or list
+    filter_half_widths: float or list
         float or list of floats of widths of delay filter windows in nanosec
```
plaplant: Should the docstring be updated to reflect the fact that it's a half-width? Also what do you mean by half-width rather than full-width?

uvtools/dspec.py
```diff
@@ -1235,9 +1241,62 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, freq_units =
...
+    assert len(wgts) == nchan, "nchan must equal length of wgts"
+    matkey = (nchan, ndelay) + tuple(wgts)
+    assert np.sum((np.abs(wgts) > 0.).astype(float)) >= ndelay, "number of unflagged channels must be greater then or equal to number of delays"
+    if fundamental_period is None:
+        fundamental_period = nchan #this tends to give well conditioned matrices. 
+    if not matkey in cache or return_diagnostics:
+        f, d = np.meshgrid(np.arange(nchan)-nchan/2, np.arange(-ndelay,ndelay), indexing='ij')
+        #f = np.asarray([[chan for m in range(ndelay)] for chan in np.arange(-nchan//2,nchan//2)])
+        #d = np.asarray([np.arange(-ndelay//2, ndelay//2) for m in range(nchan)])
+        d = d / fundamental_period
```
plaplant: This could be `d /= fundamental_period`

uvtools/dspec.py
```diff
@@ -1235,9 +1241,62 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, freq_units =
     return mdl_array, cn_array, inp_data
 
 
-def sinc_downweight_mat_inv(nchan, df, filter_centers, filter_widths,
-                            filter_factors, cache = {}, wrap = False, wrap_interval=1,
-                            nwraps = 1000, no_regularization = False):
+def delay_interpolation_matrix(nchan, ndelay, wgts, fundamental_period=None, cache={}, taper='none', return_diagnostics=False):
+    '''
+    Computes a foreground interpolation matrix that, when applied to data,
```
plaplant: When writing a docstring, the first line should be a brief imperative statement that summarizes the function's purpose. For example, this could be `Compute a foreground representation matrix for interpolating over flagged channels.`. Then after a blank line, you can have a more extended description as you do here for providing more details to the user. See [this PEP](https://www.python.org/dev/peps/pep-0257/) for more info.

uvtools/dspec.py
```diff
@@ -1235,9 +1241,62 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, freq_units =
...
+    interpolates over flagged channels with delays between -ndelays//2 and ndelays//2.
+
+    Parameters
+    ----------
+    nchan: integer
+        Number of frequency channels to interpolate over
+    ndelay,
+        number of delays to use in interpolation
+    wgts: float array
+        wgts to be applied to each frequency channel.
```
plaplant: You should note in the docstring that `wgts` needs to have length of `nchan`.

uvtools/dspec.py
plaplant: Can you also describe in the docstring a little bit how exactly you're computing this matrix? It looks like it might be a simple linear least-squares optimization, but it would be good to provide a little more of the mathematical details.

uvtools/tests/test_dspec.py
```diff
@@ -434,6 +434,20 @@ def get_snr(clean, fftax=1, avgax=0, modes=[2, 20]):
     rfft2 = np.fft.ifft2(res2)
     nt.assert_true(np.median(np.abs(rfft2[:15, :23] / rfft[:15, :23])) < 1)
 
+def test_delay_interpolation_matrix():
```
plaplant: This test looks like it tests the functionality (generating gaps in data and making sure we can interpolate over them), but it might be good to be a little more explicit about what you're testing and how. A docstring can help, as well as some more comments describing some of the main steps.

uvtools/dspec.py
anguta: I think None does make things clearer (also see below). I've changed this.

uvtools/dspec.py
anguta: I like it. Have made this change.

uvtools/dspec.py
anguta: made this change.

uvtools/dspec.py
anguta: yep. I've changed "integer" to int

uvtools/dspec.py
anguta: I've changed the assert to raising an assertionerror and also added another line to testing (to make sure its raised).

uvtools/dspec.py
anguta: changed.

uvtools/dspec.py
anguta: i've changed these to if statements that raise value errors and added tests lines.

uvtools/dspec.py
anguta: same as above.

uvtools/dspec.py
anguta: I've added this to the docstring.

uvtools/dspec.py
anguta: sure.

uvtools/dspec.py
anguta: i've changed f,d to frequencies, delays.

uvtools/dspec.py
anguta: cleaned these lines.

uvtools/dspec.py
anguta: The edge issue was primarily from when the matrices were poorly conditioned.

uvtools/dspec.py
anguta: filter_widths always meant half_widths. I've changed the variable name to accurately reflect this.

anguta: You are correct to point out that the docstring never correctly specified this. I've changed it.

uvtools/dspec.py
anguta: I've cleaned this up.

uvtools/dspec.py
anguta: I've added this to the docstring.

uvtools/dspec.py
anguta: I've added mathematical details to the docstring.

uvtools/tests/test_dspec.py
anguta: I've added a docstring with an explanation.

uvtools/dspec.py
anguta: I've changed this to a warning.

uvtools/dspec.py
```diff
@@ -1235,9 +1243,79 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, freq_units =
     return mdl_array, cn_array, inp_data
 
 
-def sinc_downweight_mat_inv(nchan, df, filter_centers, filter_widths,
-                            filter_factors, cache = {}, wrap = False, wrap_interval=1,
-                            nwraps = 1000, no_regularization = False):
+def delay_interpolation_matrix(nchan, ndelay, wgts, fundamental_period=None, cache={}, taper='none', return_diagnostics=False):
+    """
+    Copute a foreground interpolation matrix.
```
plaplant: Small typo (`Copute` -> `Compute`)

uvtools/dspec.py
```diff
@@ -1235,9 +1243,79 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, freq_units =
...
+
+    Computes a foreground interpolation matrix that, when applied to data,
+    interpolates over flagged channels with delays between
+    -ndelay / fundamental_period, ndelay / fundamental_period
+
+    The computed Matrix is equal to F = A @ [ A^T @ W @ A]^{-1} @ A^T W
```
plaplant: Fantastic explanation, thanks for adding it!

uvtools/dspec.py
```diff
@@ -1235,9 +1243,79 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, freq_units =
...
+    data, is to solve for the fourier coefficients fitting unflagged channels
+    ([ A^T @ W @ A]^{-1} @ A^T W solves the linear least squares problem) and then return
+    the unflagged Fourier transform by apply A @ to the fitted coefficients, resulting
+    in data that is linearly interpolated.
+
+    Parameters
+    ----------
+    nchan: int
+        Number of frequency channels to interpolate over.
+    ndelay,
```
plaplant: Looks like the type is still missing here (should be `ndelay: int`).

uvtools/tests/test_dspec.py
```diff
@@ -434,6 +451,48 @@ def get_snr(clean, fftax=1, avgax=0, modes=[2, 20]):
...
+    """
+    Test inerpolation matrix.
```
plaplant: Great docstring, thanks!

anguta: Thanks for pointing out where the improvement was needed ;)

uvtools/dspec.py
anguta: thanks for catching this :). I've corrected it.

uvtools/dspec.py
anguta: I've fixed this.

PR #51, opened 2019-12-03
make basemap optional. setup.py is breaking uvtools dependent packagz.

anguta:
setup.py crashes if basemap is not already installed through an alternative means (e.g. pip install --user git+https://github.com/matplotlib/basemap.git). I've made basemap optional.

### DIFF
--- a/uvtools/plot.py
+++ b/uvtools/plot.py
@@ -175,9 +175,8 @@ def plot_hmap_ortho(h, cmap='jet', mode='log', mx=None, drng=None,
     """
     try:
         from mpl_toolkits.basemap import Basemap
-    except:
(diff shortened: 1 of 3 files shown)


uvtools/plot.py
```diff
@@ -40,241 +40,244 @@ def data_mode(data, mode='abs'):
...
     -------
     plot : matplotlib.imshow
         Healpix map in ortho projection.
     """
-    from mpl_toolkits.basemap import Basemap
-    
+    try:
+        from mpl_toolkits.basemap import Basemap
+    except:
+        print("plot_hmap_ortho requires Basemap. Try running 'pip install --user git+https://github.com/matplotlib/basemap.git'")
```
philbull: This should `raise` on the line immediately afterwards.

anguta: I've converted prints to `raise ModuleNotFoundError(<helpful installation suggestion>)`

PR #54, opened 2020-01-14
Plot diff 1d

r-pascua:
This PR adds a new method to the `plot` module that allows the user to plot differences between 1d visibility arrays (or, alternatively, plot the difference between waterfalls averaged over an axis). The implementation is similar to the other `plot_diff` functions, but refactoring will be postponed until a later update.

### DIFF
--- a/uvtools/plot.py
+++ b/uvtools/plot.py
@@ -724,5 +724,125 @@ def plot_diff_uv(uvd1, uvd2, pol=None, check_metadata=True, bins=50):
 
     return fig
 
-def plot_diff_1d(uvd1, uvd2, **kwargs):
-    pass
+def plot_diff_1d(uvd1, uvd2, antpairpol, plot_type="both", 
+                 check_metadata=True, dimension="time"):
+    # TODO: docstring
+    """
+
+    """
+    if check_metadata:
+        utils.check_uvd_pair_metadata(uvd1, uvd2)
+
+    if plot_type not in ("base", "dual", "both"):
+        raise ValueError(
+            "You must specify whether to make one or two plots with "
+            "the ``plot_type`` parameter. You may choose to plot the "
+            "visibility difference as a function of frequency/time by "
+            "setting ``plot_type`` to 'base', or you can choose to "
+            "plot the difference in Fourier space by setting "
+            "``plot_type`` to 'dual'. If you would like to plot both, "
+            "then set ``plot_type`` to 'both'."
+        )
+
+    dimensions_to_duals = {"time" : "fr", "freq" : "dly"} 
+
+    if dimension not in ("time", "freq"):
+        raise ValueError(
+            "You must specify whether the visibilities are a function "
+            "of time or frequency by setting the ``dimension`` "
+            "parameter to 'time' or 'freq', respectively."
+        )
+    
+    dual = dimensions_to_duals[dimension]
+
+    use_axis = 0 if dimension == "time" else 1
+    proj_axis = (use_axis + 1) % 2
+
+    # get visibility data
+    vis1 = uvd1.get_data(antpairpol).mean(axis=proj_axis)
+    vis2 = uvd2.get_data(antpairpol).mean(axis=proj_axis)
+
+    # use same approach as in plot_diff_waterfall
+    # get important metadata
+    times = np.unique(uvd1.time_array) # days
+    lsts = np.unique(uvd1.lst_array) # radians
+    freqs = np.unique(uvd1.freq_array) # Hz
+
+    # import astropy for unit conversions
+    import astropy.units as u
+    frs = utils.fourier_freqs(times * u.day.to('s')) # Hz
+    dlys = utils.fourier_freqs(freqs) # s
+
+    # make dictionary of plotting parameters
+    plot_params = {"time" : lsts, # radians
+                   "freq" : freqs / 1e6, # MHz
+                   "fr" : frs * 1e3, # mHz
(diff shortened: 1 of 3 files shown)


uvtools/plot.py
```diff
@@ -723,3 +723,185 @@ def plot_diff_uv(uvd1, uvd2, pol=None, check_metadata=True, bins=50):
         fig.colorbar(cax)
 
     return fig
+
+def plot_diff_1d(uvd1, uvd2, antpairpol, plot_type="both", 
+                 check_metadata=True, dimension=None):
+    # TODO: docstring
```
philbull: Can be removed.

uvtools/plot.py
```diff
@@ -723,3 +723,185 @@ def plot_diff_uv(uvd1, uvd2, pol=None, check_metadata=True, bins=50):
...
+    check_metadata : bool, optional
+        Whether to check that the metadata for `uvd1` and `uvd2` match.
+        See ``utils.check_uvd_pair_metadata`` docstring for details on 
+        how the metadata are compared. If `check_metadata` is set to 
+        False, but the metadata don't agree, then the plotter may or 
+        may not error out, depending on how the metadata disagree. 
+        Default behavior is to check the metadata.
+    
+    dimension : str, optional
+        String specifying which dimension is used for the base domain. This
```
philbull: "base domain" -- could be a bit clearer what this means

uvtools/plot.py
```diff
@@ -723,3 +723,185 @@ def plot_diff_uv(uvd1, uvd2, pol=None, check_metadata=True, bins=50):
...
+
+    antpairpol : tuple
+        Tuple specifying which baseline and polarization to use to compare 
+        visibility waterfalls. See pyuvdata.UVData.get_data method docstring 
+        for information on accepted tuples.
+    
+    plot_type : str, optional
+        A string identifying which quantities to plot. Accepted values are 
+        as follows:
+            - base
```
philbull: It's hard to choose good names for these settings, but `normal` and `fourier` or similar seems a bit clearer to me.

uvtools/plot.py
```diff
@@ -723,3 +723,185 @@ def plot_diff_uv(uvd1, uvd2, pol=None, check_metadata=True, bins=50):
         fig.colorbar(cax)
 
     return fig
+
+def plot_diff_1d(uvd1, uvd2, antpairpol, plot_type="both", 
+                 check_metadata=True, dimension=None):
+    # TODO: docstring
+    """Produce plots of visibility differences along a single axis.
```
philbull: Would be good to add a note on how the averaging is done (i.e. if it takes into account weights, flags, or similar).

uvtools/plot.py
```diff
@@ -723,3 +723,185 @@ def plot_diff_uv(uvd1, uvd2, pol=None, check_metadata=True, bins=50):
...
+            "parameter to 'time' or 'freq', respectively."
+        )
+    
+    dual = dimensions_to_duals[dimension]
+
+    use_axis = 0 if dimension == "time" else 1
+    proj_axis = (use_axis + 1) % 2
+
+    # get visibility data
+    vis1 = uvd1.get_data(antpairpol).mean(axis=proj_axis)
```
philbull: I can imagine sometimes wanting to plot the median instead.

uvtools/plot.py
```diff
@@ -723,3 +723,185 @@ def plot_diff_uv(uvd1, uvd2, pol=None, check_metadata=True, bins=50):
...
+
+    # now do the same for abscissa labels
+    labels = {"time" : "LST [radians]",
+              "freq" : "Frequency [MHz]",
+              "fr" : "Fringe Rate [mHz]",
+              "dly" : "Delay [ns]"
+              }
+
+    # and now for ordinate labels
+    vis_labels = {"time" : r"$V(t)$ [Jy]",
```
philbull: Are the visibility units in the `UVData` objects guaranteed to be Jy? Can sometimes be mK.

uvtools/plot.py
r-pascua: Guess I didn't push the commit that removed this! So it's done.

uvtools/plot.py
r-pascua: Does the note that this should be either `'time'` or `'freq'` not clarify enough? Do you have a suggestion for what I could add to make it clearer?

uvtools/plot.py
r-pascua: I agree; I'll update it accordingly.

uvtools/plot.py
r-pascua: They are not; I've updated the code to use the visibility units from `uvd1`.

uvtools/plot.py
r-pascua: I've updated the code to allow the user to specify which `numpy` averaging function they would like to use, with the default being `np.mean`. This also allows the user to do a weighted average using `np.average`, so I've added `**kwargs` to the function signature, and these are passed directly to the averaging function used.

uvtools/plot.py
r-pascua: I've updated the description in the `Notes` section. With the update mentioned below, weighted averages are now allowed (with user-specified weights), but flags are not taken into account. I can make an issue to add support for flagged arrays if you think it's an important enough feature to add.

r-pascua: Something worth noting is that the addition of the `**kwargs` parameter allows users to actually retrieve the averaged visibility array by specifying the `out` parameter (although `np.average` does *not* support this).

PR #56, opened 2020-01-21
Allow tapers

r-pascua:
This PR extends the functionality of `plot.plot_diff_waterfall` and `utils.FFT` to allow a taper to be applied when taking a Fourier transform along a given axis. This supports windowing in time and frequency (in fact, `utils.FFT` should be general enough to support windowing along an arbitrary axis for an N-dimensional array).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,8 +24,8 @@ install:
   - source activate test-environment
   - conda env update -n test-environment -f environment.yml
   - pip install --user git+https://github.com/matplotlib/basemap.git # workaround
+  - pip install pyradiosky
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -25,6 +25,7 @@ install:
   - conda env update -n test-environment -f environment.yml
   - pip install --user git+https://github.com/matplotlib/basemap.git # workaround
   - pip install git+https://github.com/HERA-Team/hera_sim.git
+  - pip install git+https://github.com/RadioAstronomySoftwareGroup/pyradiosky.git
```
steven-murray, line 28: Can be switched to pypi repo.

Issue #53, opened 2019-12-27
try modeling foregrounds with DPSS 


anguta:
Apparently the eigenspectrum of the sinc matrix (with delay width \tau_w), a regularized version of which is being used in the linear filter optimizes the “centralization problem". The eigenvectors of this matrix are known as discrete prolate spheroidal sequences (DPSS) and there is a rich literature on them (as well as some convenient scipy methods).

 these eigenvectors are the set of vectors that most concentrate power within -\tau_w to \tau_w. As a result, these vectors are likely (provably) ideal for filtering/interpolating band-limited signals (isolating foregrounds in the wedge, fringe-rate filtering etc...) They might not be as flexible as the inverse sinc filter since I think they have to be defined on a regular grid (not 100% sure though) but they have the advantage of an extensive theoretical literature and are provably optimal in modeling band-limited signals. 

I don't (yet?) think they are a good choice for power-spectrum estimation 
https://github.com/HERA-Team/hera_pspec/issues/237
but I do think they are promising for foreground modeling.

anguta:
This is a nice review. https://arxiv.org/abs/1109.3649

Issue #52, opened 2019-12-17
Linear filtering with arbitrary frequencies. 


anguta:
First brought up by @AaronParsons . Linear filtering (DAYENU) and cleaning do not actually require data to exist on a regular grid. Add options for users to filter data sampled at arbitrary frequencies.

PR #57, opened 2020-01-27
Flexible interpolations

anguta:
Build in functionality for CLEAN and DFT interpolation to fit over arbitrary numbers of rectangular delay windows. 

Add methods to fit Discrete Prolate Spheroidal Sequences over arbitrary delay windows. 

Addresses issues #52 and #53 

fourier_filter method wraps up all this functionality and is designed to deprecate vis_filter, delay_filter, fringe_rate_filter, and high_pass_fourier_filter.


nkern:
@anguta can you explain what the motivation is for deprecating the other filtering methods other than just a proliferation of filtering functions? It is admirable that you compressed the functionalities into this one function, but I'm not sure if it retains all the capabilities we had before with the same clarity, but maybe I'm wrong.

uvtools/dspec.py
```diff
@@ -60,11 +60,356 @@ def calc_width(filter_size, real_delta, nsamples):
         lthresh = nsamples
     return (uthresh, lthresh)
 
-
+def fourier_filter(x, data, wgts, filter_centers, filter_half_widths, suppression_factors,
+                   mode, filter2d, fitting_options, cache=None, filter_dim=1, skip_wgt=0.1,
+                   max_contiguous_edge_flags=10):
+                   '''
+                   Your one-stop-shop for fourier filtering.
+                   We don't use the other filtering functions anymore.
```
nkern: this comment is not relevant if the other funcs are deprecated.

uvtools/dspec.py
```diff
@@ -60,11 +60,356 @@ def calc_width(filter_size, real_delta, nsamples):
...
+                   It can filter 1d or 2d data with x-axis(es) x and wgts in fourier domain
+                   rectangular windows centered at filter_centers or filter_half_widths
+                   perform filtering along any of 2 dimensions in 2d or 1d!
+                   the 'dft' and 'dayenu' modes support irregularly sampled data.
```
nkern: add line break then `Parameters`

uvtools/dspec.py
```diff
@@ -60,11 +60,356 @@ def calc_width(filter_size, real_delta, nsamples):
...
+                   -----------
+                   x: array-like
+                      Array of floats giving x-values of data. Depending on the chosen method, this data may need to be equally spaced.
+                      If performing a 2d clean, a 2-list or 2-tuple of np.ndarrays with x-values should be provided.
+                    data: array-like
+                        1d or 2d numpy.ndarray of complex data to filter.
+                    wgts: array-like
+                        1d or 2d numpy.ndarray of real weights. Must be the same shape as data.
+                    filter_centers: array-like
```
nkern: if mode is `clean` then does this need to be passed? perhaps this should be a kwarg? same for other dayenu arg inputs below

uvtools/dspec.py
```diff
@@ -60,11 +60,356 @@ def calc_width(filter_size, real_delta, nsamples):
...
+                                unstable. I don't recommend it!
+                        'dayenu_dpss_matrix', apply dayenu filter to data
+                                 and deconvolve subtracted foregrounds using
+                                 'dpss_matrix' method (see above)
+                        'dayenu_clean', apply dayenu filter to data. Deconvolve
+                                 subtracted foregrounds with 'clean'.
+                    filter2d: bool
+                        specify whether filtering will be performed in 2d or 1d.
+                        If filter is 1d, it will be applied across the -1 axis.
+                    fitting_options: dict
```
nkern: so the problem here, is that people who are using clean don't know what the default parameters are, b/c they aren't in the function signature. can't fix that given how you've implemented it here, but just FYI

uvtools/dspec.py
```diff
@@ -60,11 +60,356 @@ def calc_width(filter_size, real_delta, nsamples):
...
+                                    specifies the average degree of suppression of tones inside of the filter edges
+                                    to calculate the number of DPSS terms. Similar to edge_supression but instead checks
+                                    the suppression of a since vector with equal contributions from all tones inside of the
+                                    filter width instead of a single tone.
+                            *'clean':
+                                 'tol': float,
+                                    clean tolerance. 1e-9 is standard.
+                                 'maxiter' : int
+                                    maximum number of clean iterations. 100 is standard.
+                                 'pad': int or array-like
```
nkern: zeropad is the proper name

uvtools/dspec.py
```diff
@@ -60,11 +60,356 @@ def calc_width(filter_size, real_delta, nsamples):
...
+                                'edgecut_low' : int, number of bins to consider zero-padded at low-side of the FFT axis,
+                                    such that the windowing function smoothly approaches zero. For 2D cleaning, can
+                                    be fed as a tuple specifying edgecut_low for first and second FFT axis.
+                                'edgecut_hi' : int, number of bins to consider zero-padded at high-side of the FFT axis,
+                                    such that the windowing function smoothly approaches zero. For 2D cleaning, can
+                                    be fed as a tuple specifying edgecut_hi for first and second FFT axis.
+                                'add_clean_residual' : bool, if True, adds the CLEAN residual within the CLEAN bounds
+                                    in fourier space to the CLEAN model. Note that the residual actually returned is
+                                    not the CLEAN residual, but the residual in input data space.
+                                'taper' : window function for filtering applied to the filtered axis.
```
nkern: although `taper` is used in hera_pspec, `window` is the kwarg we have been using here. not sure if this change was accidental or not.

uvtools/dspec.py
```diff
@@ -60,11 +60,356 @@ def calc_width(filter_size, real_delta, nsamples):
...
+                        else:
+                            filter_dim_d = [1]
+                        model, residual, info = fit_basis_2d(x=x, data=data, filter_centers=filter_centers, filter_dims=filter_dim_d,
+                                                            skip_wgt=skip_wgt, basis=mode[0], method=mode[1], wgts=wgts, basis_options=fitting_options,
+                                                            filter_half_widths=filter_half_widths, suppression_factors=suppression_factors,
+                                                            cache=cache, max_contiguous_edge_flags=max_contiguous_edge_flags)
+                   elif mode[0] == 'clean':
+                        #Unpack all of the clean parameters from
+                        #fitting_options. This is to preserve default behavior
+                        #in high_pass_fourier_filter and if I had my way,
```
nkern: below comments are not relevant ;)

uvtools/dspec.py
```diff
@@ -60,11 +60,356 @@ def calc_width(filter_size, real_delta, nsamples):
...
+                        #I'd get rid of them.
+                        #However, keeping them preserves
+                        #high_pass_fourier_filter default behavior
+                        #for those who were raised on it.
+                        if not 'tol' in fitting_options:
```
nkern: I think an easier way to do this is to define a module level dictionary at the top of this file
```
CLEAN_DEFAULTS = dict(tol=1e-9, window='none', alpha=0, ...)
```
and then in this function, do one of the two, not sure which is better
```
for k in CLEAN_DEFAULTS:
  if k in fitting_options:
    locals().update(k, fitting_options[k])
  else:
    locals().update(k, CLEAN_DEFAULTS[k])
```
and then the attrs are defined in the local environment
or you can do something like
```
for k in CLEAN_DEFAULTS:
  if k not in fitting_options:
    fitting_options[k] = CLEAN_DEFAULTS[k]
```
and then just call these in this block as `fitting_options['tol']` rather than unpack them all.

Problem with both of these is that there is no typo protection, in otherwords, if I fed `fitting_options = {'tole': 1e-2}`, then the code __would not fail__, it would not see `tol` and replace it with the default and run per usual, leading to silent failure. 

The best way to resolve this, in my opinion, is to keep a separate `CLEAN` function (perhaps `high_pass_fourier_filter`?) that does have default params in its signature, and in this case you just do something like this
```
high_pass_fourier_filter(*args, **fitting_options)
```
we can discuss this next week if you like

uvtools/dspec.py
```diff
@@ -105,12 +450,14 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False,
...
             in fourier space to the CLEAN model. Note that the residual actually returned is
             not the CLEAN residual, but the residual in input data space.
-        linear : bool,
-             use aipy.deconv.clean if linear == False
-             if True, perform linear delay filtering.
+        mode : string,
+             choose from ['clean','dayenu','dft_interp']
+             use aipy.deconv.clean if 'clean'
+             use 'dayenu' if 'dayenu'
+             if 'dft_interp', then interpolates flagged channels with DFT modes.
```
nkern: does `dft_interp` belong in high_pass_fourier_filter? not sure if dft_interp is doing a high pass filter

uvtools/dspec.py
```diff
@@ -124,17 +471,31 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False,
              This allows us to avoid the problem that if we have RFI flagging and apply a linear filter
              that is larger then the horizon then the foregrounds that we fit might actually include super
              -horizon flagging side-lobes and restoring them will introduce spurious structure.
+        fg_deconv_fundamental_period: int, optional
+            fundamental period of Fourier modes to fit too.
```
nkern: not sure if I understand this based on the description. can you elaborate?

uvtools/dspec.py
```diff
@@ -190,30 +551,39 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False,
         # run clean
         if dndim == 1:
             # For 1D data array run once
-            if not linear:
+            if mode=='clean':
```
nkern: pep8: a == b. same below

uvtools/dspec.py
```diff
@@ -1237,15 +1686,490 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, freq_units =
...
 
-def delay_interpolation_matrix(nchan, ndelay, wgts, fundamental_period=None, cache={}, taper='none', return_diagnostics=False):
+def fit_basis_1d(x, y, w, filter_centers, filter_half_widths,
+                basis_options, suppression_factors=None,
+                method='leastsq', basis='dft', cache=None):
+    """
+    A 1d linear-least-squares fitting function for computing models and residuals for fitting of the form
+    y_model = A @ c
+    where A is a design matrix encoding our choice for a basis functions
+    and y_model
```
nkern: unfinished sentence

uvtools/dspec.py
```diff
@@ -1237,15 +1686,490 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, freq_units =
...
+
+    A_{nu tau} = e^{- 2 * pi * i * nu * tau / B}
+
+    for a set of taus contained within delay regions centered at filter_centers
+    and with half widths of filter_half_widths separated by 1/B where B
+    is provided by fundamental_period.
+
+    Parameters
+    ----------
+    x:
```
nkern: unfinished

uvtools/dspec.py
anguta: I think we can hold off on deprecation. I went in hoping to deprecate since (mostly my fault) high_pass_fourier_filter was turning into a mess since I tried to expand it far beyond its original intent (it was only originally supposed to support CLEAN but when I added more filtering modes, arguments exploded). I think we can decide whether we want to deprecate down the road if we find that there isn't any functionality left.

uvtools/dspec.py
anguta: thanks for catching this :)

anguta:
> @anguta can you explain what the motivation is for deprecating the other filtering methods other than just a proliferation of filtering functions? It is admirable that you compressed the functionalities into this one function, but I'm not sure if it retains all the capabilities we had before with the same clarity, but maybe I'm wrong.

The motivation for deprecating high_pass_fourier_filter was that when I added additional filtering modes, the number of arguments exploded and the code got pretty messy -- motivating me to start over. I think we can punt any deprecation decisions down the road depending on what we find ourselves actually using and not using.

uvtools/dspec.py
anguta: I've removed them ;)

uvtools/dspec.py
anguta: I think it technically could go in here since it's basically interpolating and subtracting foreground power just like `clean` but I'm also tempted to remove the all non-clean functionality from `high_pass_fourier_filter` just to clean it up. I think the function was originally intended for clean and I made the mistake of adding more and more arguments until things to really messy.

uvtools/dspec.py
anguta: if you are performing dft_interpolation, then fg_deconv_fundamental_period specifies the wavelength in the denominator of the fourier mode exponent (sets the N in e^{-2 \pi j n m / N} )

uvtools/dspec.py
anguta: Thanks for catching this. I've finished it.

uvtools/dspec.py
anguta: I've finished this one too.

uvtools/dspec.py
anguta: Should we discuss this on the analysis telecon?

uvtools/dspec.py
anguta: fixed this I think.

uvtools/dspec.py
anguta: I'm a little bit confused by your comment here. `clean` still uses `filter_centers` and `filter_half_widths` etc... to determine the areas to be cleaned in the delay transform so I do think these arguments are necessary for `clean` mode. I'm wondering if this comment is referring to something else though.

uvtools/dspec.py
anguta: That is true. I will indicate defaults in the docstring.

uvtools/dspec.py
anguta: I've mode zeropadding out of this function and will put it in `hera_cal.vis_clean` similar to how things are organized now for `vis_filter`.

uvtools/dspec.py
anguta: Technically this was an accident. Not sure which makes more sense here. Probably window. I've changed instances of taper to window.

uvtools/dspec.py
anguta: I think I can protect against typo failure by checking the keys. What do you think of this idea?

uvtools/dspec.py
```diff
@@ -1237,15 +1685,511 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, freq_units =
...
+    A 1d linear-least-squares fitting function for computing models and residuals for fitting of the form
+    y_model = A @ c
+    where A is a design matrix encoding our choice for a basis functions
+    and y_model is a fitted version of the data and c is a set of fitting coefficients determined by
+    c = [A^T w A]^{-1} A^T w y
+    where y is the original data and w is a diagonal matrix of weights for each channel in y.
+    Currently supports fitting of dpss and dft modes.
+    Parameters
+    ----------
+        x: array-like
```
nkern, line 1709: indentation here is off

uvtools/dspec.py
```diff
@@ -1237,15 +1685,511 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, freq_units =
...
+        raise ValueError("Provided 'method', '%s', is not in ['leastsq', 'matrix']."%(method))
+    model = amat @ (suppression_vector * cn_out)
+    resid = (y - model) * (np.abs(w) > 0.) #suppress flagged residuals (such as RFI)
+    return model, resid, info
+
+def fit_basis_2d(x, data, wgts, filter_centers, filter_half_widths,
+                    basis_options, suppression_factors=None,
+                    method='leastsq', basis='dft', cache=None,
+                    filter_dims = [1], skip_wgt=0.1, max_contiguous_edge_flags=5):
+        """
```
nkern, line 1828: indentation here is also off

uvtools/dspec.py
```diff
@@ -1237,15 +1685,511 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, freq_units =
...
+        A 1d linear-least-squares fitting function for computing models and residuals for fitting of the form
+        y_model = A @ c
+        where A is a design matrix encoding our choice for a basis functions
+        and y_model
+
+        Parameters:
+            x: array-like or 2-tuple/2-list
```
nkern, line 1835: indentation is doubly off I think

uvtools/dspec.py
```diff
@@ -1237,15 +1685,511 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, freq_units =
...
+                if more then one filter_dim, must provide 2-tuple or 2-list with x
+            data: array-like
+                data to fit, should be an Ntimes x Nfreqs array.
+            wgts: array-like
+                data weights.
+            filter_centers': array-like
+                list of floats specifying the centers of fourier windows with which to fit signals
+            filter_half_widths': array-like
+                list of floats specifying the half-widths of fourier windows to model.
+            suprression_factors: array-like, optional
```
nkern, line 1846: typo

uvtools/dspec.py
```diff
@@ -60,11 +67,331 @@ def calc_width(filter_size, real_delta, nsamples):
...
+                        If filter is 1d, it will be applied across the -1 axis.
+                    fitting_options: dict
+                        dictionary with options for fitting techniques.
+                        if filter2d is true, this should be a 2-tuple or 2-list
+                        of dictionaries. The dictionary for each dimension must
+                        specify the following for each fitting method.
+                            * 'dft':
+                                'fundamental_period': float or 2-tuple
+                                    The fundamental_period of dft modes to fit. This is the
+                                    Fourier resoltion of fitted fourier modes equal to
```
nkern, line 158: typo

uvtools/dspec.py
```diff
@@ -1276,68 +2222,60 @@ def delay_interpolation_matrix(nchan, ndelay, wgts, fundamental_period=None, cac
         in addition, wgts should have more nonezero values then there are
         degrees of freedom (delay modes) to solve for.
     fundamental_period: float, optional
-        the fundamental period of reconstructed delays. Default: nchan,
-        tends to give well conditioned matrices.
-        I find that 2 x nchan gives the best results (AEW).
+            fundamental period of Fourier modes to fit too.
```
nkern, line 2225: indentation is off here

uvtools/dspec.py
```diff
@@ -364,15 +736,18 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False,
...
 
     return d_mdl, d_res, info
 
-def linear_filter(data, wgts, delta_data, filter_dimensions, filter_centers, filter_half_widths, filter_factors, cache = {}):
+def dayenu_filter(x, data, wgts, filter_dimensions, filter_centers, filter_half_widths, filter_factors,
+                  cache = {}, return_matrices=True, hash_decimal=10):
     '''Apply a linear delay filter to waterfall data.
         Due to performance reasons, linear filtering only supports separable delay/fringe-rate filters.
     Arguments:
+        x: array-like or length-2 list/tuples that are array-like
```
nkern, line 749: also indented here i think

uvtools/dspec.py
nkern: got it

uvtools/dspec.py
```diff
@@ -12,6 +12,13 @@
 from scipy.optimize import leastsq, lsq_linear
 import copy
 
+#DEFAULT PARAMETERS FOR CLEANs
```
nkern, line 15: can you add a brief comment here that "True" is for 2D filtering and "False" is for 1D filtering

uvtools/dspec.py
```diff
@@ -12,6 +12,13 @@
...
+CLEAN_DEFAULTS={'tol':1e-9, 'window':{False:'none',True:['none', 'none']},
+ 'alpha':.5, 'maxiter':100, 'gain':0.1,
+ 'edgecut_low':{True:[0, 0],False:0}, 'edgecut_hi':{True:[0, 0],False:0},
+ 'add_clean_residual':False, 'filt2d_mode':'rect','add_clean_residual':False}
```
nkern, line 19: repeated key here

uvtools/dspec.py
```diff
@@ -60,11 +67,331 @@ def calc_width(filter_size, real_delta, nsamples):
...
+                                    filter window.
+                                'edge_supression': array-like
+                                    specifies the degree of supression that must occur to tones at the filter edges
+                                    to calculate the number of DPSS terms to fit in each sub-window.
+                                'avg_suppression': list of floats, optional
+                                    specifies the average degree of suppression of tones inside of the filter edges
+                                    to calculate the number of DPSS terms. Similar to edge_supression but instead checks
+                                    the suppression of a since vector with equal contributions from all tones inside of the
+                                    filter width instead of a single tone.
+                            *'clean':
```
nkern, line 184: can you add a comment that defaults can be accessed as dspec.CLEAN_DEFAULTS

uvtools/dspec.py
```diff
@@ -226,34 +581,46 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False,
...
                         else:
                             _d_cl[i] = _d[i] - _d_res[i]
                         info.append(info_here)
 
+                    elif mode == 'dft_interp':
+                        info_here = {}
+                        nmin = int((fcfg[0] - fwfg[0]) * real_delta * fg_deconv_fundamental_period[-1])
+                        nmax = int((fcfg[0] + fwfg[0]) * real_delta * fg_deconv_fundamental_period[-1])
+                        info_here['fg_deconv'] = {'method':'dft_interp','nmin':nmin, 'nmax':nmax}
+                        d_cl, _, _ = delay_filter_leastsq_1d( (data[i] * wgts[i] * win ).squeeze(), flags=(wgts[i]==0.).squeeze(), sigma=1.,
```
nkern, line 619: the conversion to flags is better done as
`flags=np.isclose(wgts[i], 0).squeeze()`
because we should avoid float-to-float equivalence

uvtools/dspec.py
```diff
@@ -226,34 +581,46 @@ def high_pass_fourier_filter(data, wgts, filter_size, real_delta, clean2d=False,
...
                                 info_fg['maxiter'] = maxiter
                                 info_fg['gain'] = gain
                                 info_here['fg_deconv'] = info_fg
                             elif fg_deconv_method == 'leastsq':
-                                nmin = int((fcfg[0] - fwfg[0]) * real_delta * data.shape[-1])
-                                nmax = int((fcfg[0] + fwfg[0]) * real_delta * data.shape[-1])
+                                nmin = int((fcfg[0] - fwfg[0]) * real_delta * fg_deconv_fundamental_period[-1])
+                                nmax = int((fcfg[0] + fwfg[0]) * real_delta * fg_deconv_fundamental_period[-1])
                                 info_here['fg_deconv'] = {'method':'leastsq','nmin':nmin, 'nmax':nmax}
                                 d_cl, _, _ = delay_filter_leastsq_1d( (data[i] * wgts[i] * win - d_r).squeeze(), flags=(wgts[i]==0.).squeeze(), sigma=1.,
```
nkern, line 607: same here

uvtools/dspec.py
```diff
@@ -1237,15 +1685,511 @@ def delay_filter_leastsq(data, flags, sigma, nmax, add_noise=False, freq_units =
...
+        float or list of floats of centers of delay filter windows in nanosec
+    filter_half_widths: float or list
+        float or list of floats of half-widths of delay filter windows in nanosec
+    cache: dictionary, optional dictionary storing operator matrices with keys
+    (x) + (filter_centers) + (filter_half_widths) + \
+    B: fundamental period of fourier modes to use for fitting. units of 1/x. For standard DFT, this is bandwidth.
+
+    Returns
+    --------
+
```
nkern, line 2159: unnecessary line breaks here

nkern:
@anguta looks fine to me overall, just some minor comments

uvtools/dspec.py
anguta: I've fixed this.

uvtools/dspec.py
anguta: I think I've fixed this?

uvtools/dspec.py
anguta: Fixed?

uvtools/dspec.py
anguta: cool. Corrected it.

uvtools/dspec.py
anguta: got it.

uvtools/dspec.py
anguta: fixed.

uvtools/dspec.py
anguta: I think i've fixed this.

uvtools/dspec.py
anguta: I've added a comment explaining the switches.

uvtools/dspec.py
anguta: I've gotten rid of the repetition.

uvtools/dspec.py
anguta: I now mention this.

uvtools/dspec.py
anguta: i've changed this.

uvtools/dspec.py
anguta: this too.

uvtools/dspec.py
anguta: got rid of the break.

anguta:
> @anguta looks fine to me overall, just some minor comments

Cool. I think I've resolved all the typos/docstring indent issues.

nkern:
@anguta looks good. This breaks some functionality in `hera_cal` (specifically the kwarg 'linear' in `vis_filter`) and also `hera_pspec` (specifically renaming of `sinc_downweight_mat_inv`). So let's get those minor PRs into hera_cal and hera_pspec as soon as you have some free time. Thanks!

PR #62, opened 2020-03-03
Fix metadata check

### LINKED ISSUE #60: utils.check_uvd_pair_metadata always fails for Ntimes=1
A ` MetadataError: "Time values disagree more than the mean integration time."` is thrown even when identical UVData objects are passed to the function.

Starting at line 181 in utils.py:
```python
dx = lambda x1, x2 : 0.5 * (np.mean(np.diff(x1)) + np.mean(np.diff(x2)))

t1vals = np.unique(uvd1.time_array)
t2vals = np.unique(uvd2.time_array)
if not np.all(
    np.isclose(t1vals, t2vals, rtol=0, atol=dx(t1vals, t2vals))
):
    raise MetadataError(
        "Time values disagree more than the mean integration time."
    )
```

if `uvd1` and `uvd2` only contain one unique time then the expression `dx(t1vals, t2vals)` always returns `nan`. [np.diff](https://docs.scipy.org/doc/numpy/reference/generated/numpy.diff.html) returns an empty array when passed an array with one element, and [np.mean](https://docs.scipy.org/doc/numpy/reference/generated/numpy.mean.html) returns nan when passed an empty array.

r-pascua:
This PR updates the `utils.check_uvd_pair_metadata` function to work properly when the data is collapsed along a single axis (i.e. only one frequency or time was used for the simulation). Note that the update is somewhat lax&mdash;if the data is collapsed along a single axis, then it assumes that the value for that axis agrees for both `UVData` objects.

This also adds some updates to the tests, to test the above claim and to ensure that the `plot.plot_diff_1d` function works if the data is collapsed along a single axis. This does not add any features to `plot.plot_diff_waterfall`, nor does it add any support for data collapsed along a single axis for this function (waterfall plots are inappropriate for 1-dimensional data). 

Closes #60.


r-pascua:
Note that this is currently failing because `pyradiosky` is not automatically installed alongside `hera_sim`, which is used for testing. This issue likely traces back to `pyuvsim` not installing `pyradiosky` automatically.

.travis.yml
```diff
@@ -24,6 +24,7 @@ install:
   - source activate test-environment
   - conda env update -n test-environment -f environment.yml
   - pip install --user git+https://github.com/matplotlib/basemap.git # workaround
+  - pip install git+https://github.com/RadioAstronomySoftwareGroup/pyradiosky.git
```
steven-murray, line 27: This can be fixed to be the pip repo now.

uvtools/utils.py
```diff
@@ -171,35 +171,39 @@ def check_uvd_pair_metadata(uvd1, uvd2):
...
 
     t1vals = np.unique(uvd1.time_array)
     t2vals = np.unique(uvd2.time_array)
-    if not np.all(
-        np.isclose(t1vals, t2vals, rtol=0, atol=dx(t1vals, t2vals))
-    ):
-        raise MetadataError(
-            "Time values disagree more than the mean integration time."
-        )
+    if not (uvd1.Ntimes == 1 or uvd2.Ntimes == 1):
```
steven-murray, line 185: I think `if uvd1.Ntimes > 1 and uvd2.Ntimes > 1` is clearer.

uvtools/utils.py
```diff
@@ -171,35 +171,39 @@ def check_uvd_pair_metadata(uvd1, uvd2):
...
+                "Time values disagree more than the mean integration time."
+            )
 
     f1vals = uvd1.freq_array[0]
     f2vals = uvd2.freq_array[0]
-    if not np.all(np.isclose(f1vals, f2vals, atol=dx(f1vals, f2vals))):
-        raise MetadataError(
-            "Frequency values disagree more than the mean channel width."
-        )
+    if not (uvd1.Nfreqs == 1 or uvd2.Nfreqs == 1):
```
steven-murray, line 195: Same comment.

PR #76, opened 2020-06-02
fix a bug in filter factors error check.

anguta:


anguta:
one line fix to an error catch.

PR #77, opened 2020-06-25
Improved waterfalls

r-pascua:
This PR adds two new plotting tools: `plot.labeled_waterfall` and `plot.fourier_transform_waterfalls`. These functions act on either `UVData` objects, provided an `antpairpol` key, or `ndarray`s with appropriate metadata (`freqs` and `lsts` or `times`) provided. `plot.labeled_waterfall` produces a single waterfall complete with axis labels, a colorbar, and colorbar label (though the colorbar may be omitted if desired), and also has support for Fourier transforming the data along any axis before plotting. Plot units are as follows: frequencies are displayed in MHz; times are displayed in JD; LSTs are displayed in hours; delays are displayed in ns; fringe-rates are displayed in mHz. Units are currently not customizable (the overhead to customize these doesn't seem worth it). `plot.fourier_transform_waterfalls` is basically a thin wrapper around `plot.labeled_waterfall`, and it plots a 2x2 grid of waterfalls, with each plot showing a different choice of Fourier transform(s).

This PR also sneaks in an update to `utils.FFT`: now it applies `np.fft.fftshift` to the (potentially tapered) data before and after the Fourier transform (before this PR, the shift was only applied after the Fourier transform).


uvtools/plot.py
```diff
@@ -471,6 +472,428 @@ def omni_view_gif(filenames, name='omni_movie.gif'):
...
+        Keyword arguments to use in generating the taper for the frequency axis.
+    time_taper: str, optional
+        Taper to use when performing a Fourier transform along the time axis.
+        Must be one of the tapers supported by :func:`dspec.gen_window`.
+    time_taper_kwargs: dict, optional
+        Keyword arguments to use in generating the taper for the time axis.
+
+    Returns
+    -------
+    fig or ax: :class:`plt.Figure` instance or :class:`plt.Axes` instance
```
dannyjacobs, line 568: Its a bit odd and potentially less than ideal to have the return type change. Since all figures have both a fig and an array of axes, why not just have the output mimic the output of  fig, axes = plt.subplots()

uvtools/plot.py
```diff
@@ -471,6 +472,428 @@ def omni_view_gif(filenames, name='omni_movie.gif'):
...
+        ``lsts`` is not provided.
+    lsts: array-like of float
+        Observed LSTs, in radians. Required if ``data`` is an array-like object
+        and ``times`` is not provided.
+    time_or_lst: str, optional
+        Either "time" or "lst". Used to specify whether the time axis should be
+        in JD or LST. If ``data`` is an array-like object and only one of ``times``
+        or ``lsts`` is provided, then this parameter is ignored.
+    mode: str, optional
+        Plotting mode to use; must be a mode accepted by :func:`data_mode`. Default
```
dannyjacobs, line 780: What is :func:data_mode? How do I get a list of options?

uvtools/plot.py
```diff
@@ -471,6 +472,428 @@ def omni_view_gif(filenames, name='omni_movie.gif'):
...
+        Dots-per-inch of the figure. Default is 100.
+    aspect: float or str, optional
+        Aspect ratio to use for each subplot. Default is "auto".
+    fontsize: float, optional
+        Font size to use for plotting labels, in points.
+    cmap: str or :class:`plt.cm.colors.Colormap` instance
+        Color map to be used when drawing waterfalls. Default is to have the choice
+        be based on the data mode selected: if ``mode`` is "phs", then "twilight"
+        is used; otherwise, "inferno" is used.
+    dynamic_range: dict, optional
```
dannyjacobs, line 794: Is there a way to set the max or min? Need one or the other to fully set the color scale.

uvtools/utils.py
```diff
@@ -115,8 +115,8 @@ def FFT(data, axis, taper=None, **kwargs):
        An array of data, assumed to not be ordered in the numpy FFT convention.
        Typically the data_array of a UVData object.
 
-    axis : int
-        The axis to perform the FFT over.
+    axis : int, optional
```
dannyjacobs, line 118: Assuming this means for now we don't support delay and fringe rate at the same time?

uvtools/plot.py
```diff
@@ -471,6 +472,428 @@ def omni_view_gif(filenames, name='omni_movie.gif'):
...
+
+    # Update data for plotting.
+    data = data_mode(data, mode=mode)
+    vmin, vmax = None, None
+
+    # Prepare colorbar parameters.
+    if mode == "phs":
+        cbar_label = "Phase [radians]"
+        vmin, vmax = -np.pi, np.pi
+        if cmap == "best":
```
dannyjacobs, line 649: I've always wanted to find a wrapping colorscale for phase. Every once in a while I check to see if matplotlib or somewhere else has one.  Is Twilight such a thing? If not could you check for me? Someday my dream will come true.

uvtools/plot.py
r-pascua: Noted. It seemed straightforward to me that passing an `Axes` object (by setting `ax=...` in the function call) would get an `Axes` object returned, and a figure otherwise, though I do see how it can be confusing. I can go ahead and make the change.

uvtools/plot.py
r-pascua: Ah, that type of docstring formatting is intended to produce a link to the function in the compiled `rst` version of the documentation (so on the readthedocs). `data_mode` is the first function defined in the `plot` module.

uvtools/plot.py
r-pascua: It's currently handled under the hood. I can add extra options for controlling that at the function call, though it may have unexpected side effects if users give `vmin`/`vmax` parameters that are incompatible with the `mode` being used.

uvtools/utils.py
r-pascua: That's right; to do both you would need to do two successive calls of `FFT`. If you'd like it changed, then I can change it, but it'll have a few downstream consequences I'll have to patch up.

uvtools/plot.py
r-pascua: "twilight" is indeed a wrapping colorscale. It goes black -> blue -> white -> red -> black (either that, or reversed). There's also "twilight_r", "twilight_shifted" and "hsv". This is a good reference for colormaps: https://matplotlib.org/3.1.1/gallery/color/colormap_reference.html

uvtools/utils.py
r-pascua: I figure I should note that the intention here is that `FFT` is a convenience function for performing a FFT, with optional tapering, that handles the shifting under the hood. I was mainly fed up with having to do `np.fft.fftshift(np.fft.fft(np.fft.fftshift(...)))` whenever I wanted to Fourier transform something, and wanted to have the option to apply a taper without doing the extra work of generating the window and so on.

r-pascua:
Thanks for the review! I'll implement the suggested changes and update the tests shortly.

r-pascua:
OK, I think I've addressed all the concerns raised in the comments, and I've updated the tests appropriately.

uvtools/plot.py
dannyjacobs: you need two parameters to set the color scale, either max and min, or max and drange.  Here you've already got drange so just add a max.  If None, let the default float to max.

uvtools/utils.py
dannyjacobs: Are things flexible enough to add 2D FFT down the road without changing interfaces much? If so, then its fine to file an issue and leave it for the future.

r-pascua: It's possible. My preferred way of doing it would be to make a thin wrapper `FFT2D`, rather than make the code in `FFT` messy, but I suppose that details can be discussed later

uvtools/plot.py
r-pascua: So I've added `vmin` and `vmax` as optional parameters, and the color scale normalization is calculated using whichever two of (`vmin`, `vmax`, `dynamic_range`) are specified. If all three are provided, then it just uses `vmin` and `vmax`.

r-pascua: I should make it clear that I added this functionality for `labeled_waterfall`, but not for `fourier_transform_waterfalls`, since the latter would require extra handling (it doesn't make sense to apply the same `vmin` or `vmax` parameter to two sets of data if one of them has been Fourier transformed). If `dynamic_range` is the only parameter provided, then it calculates the max under the hood and sets the minimum value according to that and the dynamic range.

dannyjacobs: Ok. Lets roll with this and people can propose changes as they play.

uvtools/utils.py
dannyjacobs: Ok. Please open an issue to put down a marker.

PR #88, opened 2020-07-08
Fix fft normalization

r-pascua:
This PR fixes the normalization of the FFT data in `plot.labeled_waterfall` and the functions that wrap it (currently just `plot.fourier_transform_waterfalls`). Previously, the FFT data only used the shifted output from `np.fft.fft`; now, the data is scaled by the resolution of the axis that the transform is performed over, in order to ensure correct units of transformed data.


r-pascua:
Closing to reopen so that tests can re-run.

PR #91, opened 2020-07-31
Set flagged residuals to zero

anguta:
small change to `dayenu_filter` that ensures that flagged data in the residual is equal to zero.


anguta:
Thank you @r-pascua  for the quick review! I've added a 2d check. It passes on my computer so I think it's ready for another look but it looks like something is broken  in travis. Let's make this a different issue and PR.

r-pascua:
Ok, happy to approve since you say its passing locally. Definitely agree that the `Basemap` issue should be handled in a different PR.

PR #96, opened 2020-08-05
Rect 2d filter fit basis

### LINKED ISSUE #95: add `rect` 2d filtering mode to `_fit_basis_2d`

anguta:
closes #95


uvtools/dspec.py
```diff
@@ -477,6 +498,7 @@ def fourier_filter(x, data, wgts, filter_centers, filter_half_widths, mode,
                            filter_dims_d = [1]
                        suppression_factors = filter_kwargs.pop('suppression_factors')
                        max_contiguous_edge_flags = filter_kwargs.pop('max_contiguous_edge_flags')
+                       filt2d_mode = filter_kwargs.pop('filt2d_mode')
```
r-pascua, line 501: `filt2d_mode` is popped from `filter_kwargs` here, but it doesn't look like it's passed to `_fit_basis_2d`&mdash;is this intentional?

uvtools/dspec.py
```diff
@@ -1661,29 +1685,40 @@ def _clean_filter(x, data, wgts, filter_centers, filter_half_widths,
...
-                if fc0 >= _x[0].min() and fc0 <= _x[0].max():
+        # in rectangle mode, the list of time filter centers / widths and frequency filter centers / widths
+        # specify rectangular regions in 2d Fourier space to CLEAN in. This loop iterates over those centers
+        # and widths.
+        for fc_t, fw_t, fc_f, fw_f in zip(filter_centers[0], filter_half_widths[0], filter_centers[1], filter_half_widths[1]):
+            if filt2d_mode == 'rect':
+                # determine rectangular region by taking outer products of filter areas along fourier time and fourier freq
+                # axes.
+                area_t = np.outer(_get_filter_area(x[0], fc_t, fw_t), _get_filter_area(x[1], fc_f, fw_f))
+            elif filt2d_mode == 'plus':
```
r-pascua, line 1708: Inconsistency here with what's written in the docstring: docstring says that supported modes are `'rect'` and `'outer'`, but this uses `'rect'` and `'plus'`.

uvtools/dspec.py
```diff
@@ -1661,29 +1685,40 @@ def _clean_filter(x, data, wgts, filter_centers, filter_half_widths,
...
+                    area_t[np.argmin(np.abs(_x[0]-fc_t)), :] = av
+                if fc_f >= _x[1].min() and fc_f <= _x[1].max():
                     #generate area vector centered at zero
-                    av = _get_filter_area(x[0], fc0, fw0)
-                    area_temp[:, np.argmin(np.abs(_x[1]-fc1))] = av
-                area += area_temp
+                    av = _get_filter_area(x[0], fc_t, fw_t)
+                    area_t[:, np.argmin(np.abs(_x[1]-fc_f))] = av
+            else:
+                raise ValueError("%s is not a valid filt2d_mode! choose from ['rect', 'plus']"%(filt2d_mode))
```
r-pascua, line 1719: Can write this a bit more compactly using an f-string: `f"{filt2d_mode} is not a valid filt2_mode! ..."`

uvtools/dspec.py
```diff
@@ -1876,52 +1920,109 @@ def _fit_basis_2d(x, data, wgts, filter_centers, filter_half_widths,
...
-        info['filter_params']['axis_1']['basis_options'] = info_t['basis_options']
-        info['filter_params']['axis_1']['mode'] = info_t['basis'] + '_' + method
-    if filter2d:
-        wgts_time = np.ones_like(wgts)
-        for i in range(data.shape[0]):
-            if info['status']['axis_1'][i] == 'skipped':
-                wgts_time[i] = 0.
-        for i, _y, _w, in zip(range(model.shape[1]), model.T, wgts_time.T):
+    if filt2d_mode == 'outer' or not filter2d:
+        for i, _y, _w, in zip(range(data.shape[0]), data, wgts):
```
r-pascua, line 1924: An explanatory comment would be helpful here (it's a little hard to parse whether the diff makes sense without knowing what this is meant to do).

uvtools/dspec.py
```diff
@@ -1876,52 +1920,109 @@ def _fit_basis_2d(x, data, wgts, filter_centers, filter_half_widths,
...
+                    info['filter_params']['axis_0']['round%d'%stepnum]['filter_half_widths'] = info_t['filter_half_widths']
+                    info['filter_params']['axis_0']['round%d'%stepnum]['suppression_factors'] = info_t['suppression_factors']
+                    info['filter_params']['axis_0']['round%d'%stepnum]['basis_options'] = info_t['basis_options']
+            # and subtract it (effectively subtracting rectangular Fourier model)
+            residual = (residual - model_temp) * (np.abs(wgts) > 0.).astype(float)
+            # and add the model to the overall model.
+            model += model_temp
+            # repeate this for all the rectangle bounds provided.
+    else:
+        raise ValueError("%s not a supported filt2d_mode ('rect', 'outer')"%filt2d_mode)
```
r-pascua, line 2025: Another place where an f-string could be used instead of the old style string formatting.

uvtools/tests/test_dspec.py
```diff
@@ -890,6 +890,24 @@ def get_snr(clean, fftax=1, avgax=0, modes=[2, 20]):
...
+    mdl1, res1, info1 = dspec.fourier_filter(x=[times, freqs], data=d, wgts=w, filter_centers=[[0.],[0.]],
+                                             filter_half_widths=[[fr_len],[bl_len]], suppression_factors=[[1e-8],[1e-8]],
+                                             mode='dayenu_dpss_leastsq', filter_dims=[1, 0], **dpss_options1_2d)
+    mdl2, res2, info2 = dspec.fourier_filter(x=[times, freqs], data=d, wgts=w, filter_centers=[[0.],[0.]],
+                                             filter_half_widths=[[fr_len],[bl_len]], suppression_factors=[[1e-8],[1e-8]],
+                                             mode='dayenu_dpss_leastsq', filter_dims=[1, 0], **dpss_options2_2d)
+    nt.assert_true(np.all(np.isclose(mdl1, mdl2)))
+    nt.assert_true(np.all(np.isclose(res1, res2)))
+    # assert ValueError for unsupported filt2d_mode
+    nt.assert_raises(ValueError, dspec.fourier_filter, x=[times, freqs], data=d, wgts=w, filter_centers=[[0.],[0.]],
```
r-pascua, line 907: If possible, it would be good to run similar tests with the other filtering modes. In particular, I would expect this same set of tests to fail if it were run using something that calls `_clean_filter`. On that note, it looks like only the `'clean'` mode will call this, but `'dayenu_clean'` will not&mdash;is that intended?

uvtools/tests/test_dspec.py
```diff
@@ -890,6 +890,24 @@ def get_snr(clean, fftax=1, avgax=0, modes=[2, 20]):
     nt.assert_raises(ValueError, dspec.fourier_filter,x=flog, data=d, wgts=w, filter_centers=[0.],
                                                   filter_half_widths=[bl_len],
                                                   mode='clean', filter_dims=[1], **{'tol':1e-5})
+
+    # check that 2d clean with single rectangular region is the same for outer and rect filter2d modes.
+    dpss_options1_2d = {'eigenval_cutoff': [[1e-12], [1e-12]], 'filt2d_mode': 'rect'}
+    dpss_options2_2d = {'eigenval_cutoff': [[1e-12], [1e-12]], 'filt2d_mode': 'outer'}
+    dpss_options3_2d = {'eigenval_cutoff': [[1e-12], [1e-12]], 'filt2d_mode': 'plus'}
+    mdl1, res1, info1 = dspec.fourier_filter(x=[times, freqs], data=d, wgts=w, filter_centers=[[0.],[0.]],
```
r-pascua, line 898: I'm not super familiar with what to expect from this, but it seems like choosing the origin as the filter center could be too simple of a test? At the very least, it would be nice to see that this works for any choice of filter center. (I trust that you've convinced yourself that it will work, but it would be reassuring to see this be tested at a non-trivial choice of filter center if possible.)

uvtools/tests/test_dspec.py
```diff
@@ -890,6 +890,24 @@ def get_snr(clean, fftax=1, avgax=0, modes=[2, 20]):
     nt.assert_raises(ValueError, dspec.fourier_filter,x=flog, data=d, wgts=w, filter_centers=[0.],
                                                   filter_half_widths=[bl_len],
                                                   mode='clean', filter_dims=[1], **{'tol':1e-5})
+
+    # check that 2d clean with single rectangular region is the same for outer and rect filter2d modes.
```
r-pascua, line 894: Can you make this its own test? I think best practice is to use directed, atomic tests with as few failure points as possible, rather than large tests that have many failure points.

PR #102, opened 2020-08-08
Fix dayenu resid

anguta:
changing the syntax in `dayenu_filter` on the line where `residual` causes unittests that ensure the zero was set properly in `hera_cal` to pass https://github.com/HERA-Team/hera_cal/pull/619.

 I really don't know why.

I also added a line that stores `filter_factors` in the `info` dictionary returned by `dayenu_filter`. It was not being set before and should have been.

### DIFF
--- a/uvtools/dspec.py
+++ b/uvtools/dspec.py
@@ -450,7 +450,7 @@ def fourier_filter(x, data, wgts, filter_centers, filter_half_widths, mode,
                                                      filter_centers=filter_centers, filter_half_widths=filter_half_widths,
                                                      filter_factors=suppression_factors, cache=cache, skip_wgt=skip_wgt,
                                                      max_contiguous_edge_flags=max_contiguous_edge_flags)
-                       model = data - residual * (np.abs(wgts) > 0.).astype(float)
+                       model = data - residual
                        if len(mode) > 1:
                            model, _, info_deconv = _fit_basis_2d(x=x, data=model, filter_centers=filter_centers, filter_dims=filter_dims_d,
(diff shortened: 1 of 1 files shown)


nkern:
I think this is because the float comparison isn't actually working. anytime you need to compare floats (unless you are checking that they are the same variable, which isn't the case here) you should use `np.isclose`. I would recommend using that for now

anguta:
> I think this is because the float comparison isn't actually working. anytime you need to compare floats (unless you are checking that they are the same variable, which isn't the case here) you should use `np.isclose`. I would recommend using that for now

@nkern thanks for pointing this out! I've changed the lines to use `isclose` instead.

nkern:
@anguta I remember now why the residual is not actually zero at the flagged channels, this is because the residual is defined in Fourier space, such that the residual in data space is the invFT(resid). Depending on the filter, I believe this can lead to non-zero resid in data space in the flagged channels.

I recall doing some work a while back when @jsdillon and I were working on high_pass_fourier_filter on this. I know that this PR is mainly for setting the resid and model to zero at skipped integrations, so I'm fine with that. But can you comment on why setting the residual to zero in the flagged channels is mandated here? I would prefer to actually keep this functionality (resid is possibly non-zero at flagged channels). thanks

anguta:
@nkern , I've added an option to let you turn zeroing flags on and off. 

The main reason to set the resid to zero in flagged channels is that if there is in fact RFI in this channel, then the resid will have RFI in it. If you try add this residual to the filtered model to get cleaned data, then the cleaned data will have the model plus RFI when in fact you actually want the cleaned data to have the model inside of the flagged channels.

nkern:
@anguta cool, thanks

PR #103, opened 2020-09-26
Minor python3 fixes to plot_uv.py

AaronParsons:
plot_uv.py is super legacy, but sometimes helpful. This just makes it work again.


anguta (approved):
Looks good to me! Thanks for keeping this up to date :)

PR #106, opened 2020-12-03
fix fundamental period bug.

anguta:
Fixes a bug where the fundamental period for DFT fits was getting inadvertently set to zero on 1d filtering.

### DIFF
--- a/uvtools/tests/test_dspec.py
+++ b/uvtools/tests/test_dspec.py
@@ -808,7 +808,8 @@ def test_fourier_filter():
     #check that fringe rate filter model gives similar results to delay filter.
     nt.assert_true(np.all(np.isclose(mdl1[~f],mdl5[~f], rtol=1e-2)))
 
-    #perform some sanity checks on handling of nans in dft_leastsq
+    #perform some sanity checks on handling of nans in dft_leastsq. If nans are present in fundamental period
+    #then the default behavior should be to set fundamental period to 2 * bandwidth.
     mdl, res, info = dspec.fourier_filter(x=freqs, data=d[0], wgts=w[0], filter_centers=[0.],
                                              filter_half_widths=[bl_len], suppression_factors=[1e-9],
                                              mode='dft_leastsq')
--- a/uvtools/tests/test_dspec.py
+++ b/uvtools/tests/test_dspec.py
@@ -784,7 +784,6 @@ def test_fourier_filter():
     mdl11d, res11d, info11d = dspec.fourier_filter(x=freqs, data=d[0], wgts=w[0], filter_centers=[0.],
                                              filter_half_widths=[bl_len], suppression_factors=[0.],
                                              mode='dpss_leastsq', **dpss_options1)
-    #perform some sanity checks on handling of nans in dft_leastsq
 
     #test that the info is properly switched. fourier_filter processes all data in frequency_mode and takes transposes for time
     #filtering mode.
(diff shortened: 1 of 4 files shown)


r-pascua (commented):
Are you sure that this is not going to introduce a new bug? It looks like this fix will result in 1d dft-based filters using `2 * (x.max() - x.min())` as the fundamental period regardless of what the user passes as the fundamental period (in particular, I'm looking at the two lines immediately following where you made changes). You can probably fix this by doing `fp = [...]` only if `np.isnan(fp[0])`. 

On a related note, I'd suggest thinking about writing some tests that would cover this sort of thing. Perhaps something that does sanity checks for things like this (for example, the test could parse the info dictionary and check that the stuff in there agrees with properties of the filtered data as well as the arguments the user provided).

anguta:
> Are you sure that this is not going to introduce a new bug? It looks like this fix will result in 1d dft-based filters using `2 * (x.max() - x.min())` as the fundamental period regardless of what the user passes as the fundamental period (in particular, I'm looking at the two lines immediately following where you made changes). You can probably fix this by doing `fp = [...]` only if `np.isnan(fp[0])`.
> 
> On a related note, I'd suggest thinking about writing some tests that would cover this sort of thing. Perhaps something that does sanity checks for things like this (for example, the test could parse the info dictionary and check that the stuff in there agrees with properties of the filtered data as well as the arguments the user provided).

Thanks for the catch! I've fixed this and I'll add some tests.

anguta:
Sorry for taking so long to get to this unittest. I've finally added it. Let me know if this is sufficient.

uvtools/tests/test_dspec.py
```diff
@@ -774,6 +784,7 @@ def get_snr(clean, fftax=1, avgax=0, modes=[2, 20]):
     mdl11d, res11d, info11d = dspec.fourier_filter(x=freqs, data=d[0], wgts=w[0], filter_centers=[0.],
                                              filter_half_widths=[bl_len], suppression_factors=[0.],
                                              mode='dpss_leastsq', **dpss_options1)
+    #perform some sanity checks on handling of nans in dft_leastsq
```
r-pascua, line 787: Comment is in the wrong place.

uvtools/tests/test_dspec.py
```diff
@@ -797,6 +808,17 @@ def get_snr(clean, fftax=1, avgax=0, modes=[2, 20]):
 
     #check that fringe rate filter model gives similar results to delay filter.
     nt.assert_true(np.all(np.isclose(mdl1[~f],mdl5[~f], rtol=1e-2)))
+
+    #perform some sanity checks on handling of nans in dft_leastsq
+    mdl, res, info = dspec.fourier_filter(x=freqs, data=d[0], wgts=w[0], filter_centers=[0.],
```
r-pascua, line 813: I think it could be made a little more clear that this is testing that nans are handled correctly. I understand that the default setting for a dft-based filter is to set the fundamental period to `np.nan`, but this may be confusing if there are other people who contribute to these tools in the future. I suppose this isn't essential, but would be good to implement in the future if we're planning on raising the standards to which we hold the code in this repo.

uvtools/tests/test_dspec.py
```diff
@@ -797,6 +808,17 @@ def get_snr(clean, fftax=1, avgax=0, modes=[2, 20]):
...
+                                             filter_half_widths=[bl_len], suppression_factors=[1e-9],
+                                             mode='dft_leastsq')
+    # check that the filter_period is indeed equal to 1 / (2 * bandwidth)
+    nt.assert_true(np.isclose(info['filter_params']['axis_1']['basis_options']['fundamental_period'],
```
r-pascua, line 817: This doesn't test all cases of `dft_leastsq` handling the fundamental period. It tests that the default fundamental period is set appropriately, but it doesn't test the case where the user provides a fundamental period different than the default setting. This second part is important to test, since this test would *not* catch the bug I pointed out in a previous comment (where the code was setting the fundamental period to the default regardless of the input it was provided).

r-pascua (changes requested):
Glad to see more tests added, but please see the last comment left on my review. I think it's important to address that point, since it has to do with ensuring we catch potential silent bugs.

uvtools/tests/test_dspec.py
anguta: fixed.

uvtools/tests/test_dspec.py
anguta: I've added a test. Also check 2d.

uvtools/tests/test_dspec.py
anguta: I've added additional documentation to explain this.

anguta:
I've added unittests checking that fundamental period is correctly added to the info dictionary and additional documentation.

r-pascua (approved):
Looks good to me! Thanks for making the requested changes!

PR #110, opened 2021-01-16
Phase wraps

r-pascua:
This PR fixes a silent bug in `plot.labeled_waterfall`: prior to this update, the LST axis ticklabels were incorrect when the provided LSTs were wrapped. For example, if the LST array started at 23 hours and ended at 5 hours, then the plotting function would have listed the LST values as ranging from 5 hours to 23 hours, rather than wrapping from 23 hours to 5 hours--this update fixes that problem.


uvtools/plot.py
```diff
@@ -469,6 +469,7 @@ def labeled_waterfall(
     vmin=None,
     vmax=None,
     dynamic_range=None,
+    Nticks=6,
```
steven-murray, line 472: Let's try to use standard python formatting for variables


```suggestion
    n_ticks=6,
```

uvtools/plot.py
```diff
@@ -594,6 +612,15 @@ def labeled_waterfall(
             raise TypeError("array-like data must consist of complex numbers.")
         if data.ndim != 2 or (data.ndim == 2 and 1 in data.shape):
             raise ValueError("array-like data must be 2-dimensional.")
+    if type(Nticks) is not int:
```
steven-murray, line 615:
```suggestion
    if not isinstance(n_ticks, int):
```

steven-murray (commented):
Thanks @r-pascua, maybe we should get this merged 😆

PR #111, opened 2021-03-22
Dpss xtol

aewallwi:
short PR that addresses the fact that the time-grids in lst-binned files can have some jitter so they are not equally spaced within the default np.isclose tolerance. Adds an optional atol parameter to the dpss_operator method that is by default 1/10 the average spacing.

### DIFF
--- a/uvtools/dspec.py
+++ b/uvtools/dspec.py
@@ -1999,7 +1999,7 @@ def fit_solution_matrix(weights, design_matrix, cache=None, hash_decimal=10, fit
 
 def dpss_operator(x, filter_centers, filter_half_widths, cache=None, eigenval_cutoff=None,
         edge_suppression=None, nterms=None, avg_suppression=None, xc=None, hash_decimal=10,
-        xtol=0.1):
+        xtol=1e-3):
     """
     Calculates DPSS operator with multiple delay windows to fit data. Frequencies
     must be equally spaced (unlike Fourier operator). Users can specify how the
--- a/uvtools/dspec.py
+++ b/uvtools/dspec.py
@@ -2066,7 +2066,7 @@ def dpss_operator(x, filter_centers, filter_half_widths, cache=None, eigenval_cu
                                  label='dpss_operator', crit_val=tuple(crit_provided_value[0]))
     if not opkey in cache:
         #check that xs are equally spaced.
-        if not np.all(np.isclose(np.diff(x), np.mean(np.diff(x))), rtol=0., atol=xtol * np.mean(np.diff(x))):
+        if not np.all(np.isclose(np.diff(x), np.mean(np.diff(x)), rtol=0., atol=xtol * np.mean(np.diff(x)))):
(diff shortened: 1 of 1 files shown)


uvtools/dspec.py
```diff
@@ -2062,7 +2066,7 @@ def dpss_operator(x, filter_centers, filter_half_widths, cache=None, eigenval_cu
                                  label='dpss_operator', crit_val=tuple(crit_provided_value[0]))
     if not opkey in cache:
         #check that xs are equally spaced.
-        if not np.all(np.isclose(np.diff(x), np.mean(np.diff(x)))):
+        if not np.all(np.isclose(np.diff(x), np.mean(np.diff(x)), rtol=0., atol=xtol * np.mean(np.diff(x)))):
```
jsdillon, line 2069: There should probably be an absolute value in atol to handle the case where `x` is decreasing.

uvtools/dspec.py
```diff
@@ -1998,7 +1998,8 @@ def fit_solution_matrix(weights, design_matrix, cache=None, hash_decimal=10, fit
 
 
 def dpss_operator(x, filter_centers, filter_half_widths, cache=None, eigenval_cutoff=None,
-        edge_suppression=None, nterms=None, avg_suppression=None, xc=None, hash_decimal=10):
+        edge_suppression=None, nterms=None, avg_suppression=None, xc=None, hash_decimal=10,
+        xtol=0.1):
```
jsdillon, line 2002: this feels like a very high xtol for default. I'd recommend making it closer to machine precision (1e-6 or less) and forcing the user to acknowledge to realize there's something weird and deal with it.

jsdillon (changes requested):

uvtools/dspec.py
aewallwi: I've changed this to `1e-3` which is what we see out of the lstbinner

https://github.com/HERA-Team/hera_cal/issues/698

uvtools/dspec.py
aewallwi: good point. I've adopted this.

uvtools/dspec.py
```diff
@@ -2033,6 +2034,9 @@ def dpss_operator(x, filter_centers, filter_half_widths, cache=None, eigenval_cu
         all tones inside of the filter width instead of a single tone.
     xc: float optional
     hash_decimal: number of decimals to round for floating point dict keys.
+    xtol: fraction of average diff that the diff between all x-values must be within
+          the average diff to be considered
+          equally spaced. Default is 0.1
```
jsdillon, line 2039: docstring is out of date here

jsdillon (dismissed):
One docstring is out of date, but otherwise good to go.

uvtools/dspec.py
aewallwi: fixed.

jsdillon (approved):

PR #114, opened 2021-04-22
Pytest, github-actions, and codecov.

aewallwi:
Migrate unittests from nosetest to pytest and add github actions and codecov uploading.


philbull (approved):
Looks good to me, and the new tests seem to run fine. I'd consider removing Python 3.7 from the build, as running the tests against 3 Python versions each time seems slightly excessive.

PR #116, opened 2021-05-01
Fix pytest

aewallwi:
Fix lingering issues in pytest migration that are hindering proper coverage recording and codecov checks. Verbatum translation of existing unittests into pytest compatible tests. Issues of organization of the existing tests addressed in #117.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,7 @@
+[run]
+omit =
+     */tests/*
+
+[report]
+omit =
+     */tests/*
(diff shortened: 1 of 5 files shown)


r-pascua (approved):
It looks like the coverage is working and that all the tests are passing. Since we're not worrying about modularizing tests in this PR, I'm happy to approve as-is. I think that now the tests for the plotting tools I wrote are kind of wonky, but I'll get to that in a future PR.

Thank you for doing this!

PR #128, opened 2021-05-11
Repair dpss grid

aewallwi:
Incorporate `place_data_on_uniform_grid()` into `dpss_operator` so that we can generate DPSS fitting operators on data that is on a grid but the grid has missing values (for example, if we want to jointly fit multiple disjoint spectral windows)


jsdillon (approved):
Looks fine to me. For posterity, you might also want explain in the PR (or maybe the comments or both) what motivated the choice of turning errors into skips. Is ever a cause for concern?

PR #130, opened 2021-06-01
make aipy optional (only required for clean).

aewallwi:
Make aipy purely optional.


jsdillon (approved):
Looks reasonable, assuming tests pass

PR #135, opened 2021-09-15
Dont repeate dpss window calc

aewallwi:
Eliminate wasteful repeated calculation of DPSS vectors in filtering method.


uvtools/dspec.py
```diff
@@ -2253,9 +2255,9 @@ def dft_operator(x, filter_centers, filter_half_widths,
     if cache is None:
         cache = {}
     #if no fundamental fourier period is provided, set fundamental period equal to measurement
-    #bandwidth.
+    #bandwidth * 2.
     if fundamental_period is None:
-        fundamental_period = np.median(np.diff(x)) * len(x)
+        fundamental_period = np.median(np.diff(x)) * len(x) * 2
```
jsdillon, line 2260: I don't understand this change well enough to comment on it, so it's a bit concerning that it's not covered by unit tests.

jsdillon (changes requested):
Jut one thing that I think should be unit tested and isn't currently.

PR #138, opened 2021-09-17
get rid of hera_sim requirements.

### LINKED ISSUE #136: pyuvsim API updates are breaking the plot tests.

aewallwi:
Remove dependencies on hera_sim and pyuvsim which are only used to set up tests for the plotting module -- tests that can  be satisfied using UVData files. I have created uvdata files for this purpose and they replace the `hera_sim` calls in `test_plot`.

This is in the spirit of keeping `uvtools` as a low-level module. In contrast, `hera_sim` is high level and has a large number of dependancies.

closes #136 

This was mainly a reaction to tests failing because of `hera_sim` errors that appear to arise from a break in the ways that the latest versions of `hera_sim`and `pyuvsim` interact with eachother. I think it makes sense for this module to stay independent of this sort of issue, especially since its use of `hera_sim` is not central to the modules functionality (`hera_sim` is simply used to set up test data which can be done offline).


steven-murray (approved):
This looks very sensible to me!

PR #141, opened 2022-03-10
Improve the speed of the sinc matrix eigenvalue calculation by using dpss_operator

tyler-a-cox:
When calculating eigenvalues in `uvtools.dspec.dpss_operator` to determine the number of DPSS vectors used when filtering data, we currently are recalculating the eigenvalues from the sinc matrix instead of using the values computed by `scipy.signal.windows.dpss`. This fix should speed up the runtime of this function without changing the values of the eigenvalues produced.

### DIFF
--- a/uvtools/dspec.py
+++ b/uvtools/dspec.py
@@ -2185,10 +2185,13 @@ def dpss_operator(x, filter_centers, filter_half_widths, cache=None, eigenval_cu
         if nterms is None:
             nterms = []
             for fn,fw in enumerate(filter_half_widths):
-                dpss_vectors, eigvals = windows.dpss(nf, nf * df * fw, nf, return_ratios=True)
                 if not eigenval_cutoff is None:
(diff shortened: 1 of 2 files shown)


tyler-a-cox:
I ended up changing this branch slightly. Now, I'm computing the number of eigenvalues before performing the matrix decomposition by adding the number of eigenvalues close to unity (Slepian 1978) with the number of eigenvalues that fall between the interval eps < eigenvalue < 1 - eps, where eps is the eigenvalue cutoff in `dpss_operator` (Karnik 2020; Theorem 1). By computing the number of eigenvalues we need before computing the eigenvectors, we reduce the computation time by quite a bit. Here's a comparison:

Screen Shot 2022-05-11 at 12 15 13 PM

This works especially well for large grids:

Screen Shot 2022-05-11 at 12 38 25 PM

aewallwi:
Looks good to me.

aewallwi (approved):
Looks good and great speedup of the filters.

PR #143, opened 2022-05-12
make uvtools filter only.

aewallwi:
This PR removes the utils and plotting functionality which are not really used in the data processing pipes. I think it would be a good idea to split these into their own repos or perhaps split the filtering machinery into it's own repo since the latter is frequently used by the pipes and it would help better delegate responsibility for maintaining the utils and plots modules.


bhazelton:
I was just pushing Dara to move some of the utility and plotting code into this repo from the notebook template repo because it is more widely useful. Since it is starting to be used outside the nightly notebooks it needs to be in a repo where it can be properly reviewed and tested as collaboration code. My understanding was that this was the correct repo for generally useful utility and plotting code.

aewallwi:
> I was just pushing Dara to move some of the utility and plotting code into this repo from the notebook template repo because it is more widely useful. Since it is starting to be used outside the nightly notebooks it needs to be in a repo where it can be properly reviewed and tested as collaboration code. My understanding was that this was the correct repo for generally useful utility and plotting code.

I see. Perhaps it would make sense for us to move the filter code to a separate repo then.

jsdillon:
If we do move it to a separate repo, can we have a legacy module here that just does `from hera_filters.dspec import *`... perhaps with a deprecation warning? It will certainly ease the transition.

PR #144, opened 2022-05-18
Move filtering code into hera_filters.

aewallwi:
Move filtering code out of `uvtools` into the `hera_filters` repository. `hera_filters` is at https://github.com/HERA-Team/hera_filters . The reason for this change is that foreground filtering is widely applicable beyond HERA and it makes sense to decouple it from the visualization functionality in `uvtools` and the associated dependancies. We may want to change the name of `uvtools` to `hera_visualization` if we decide to add substantially more plotting functionality in the future.

For now, I've preserved `uvtools.dspec` and just import `hera_filters` methods. In a future PR (after people have had time to migrate downstream code) we should probably completely eliminate `uvtools.dspec`.


steven-murray:
@jsdillon @tyler-a-cox @aewallwi I think someone should take over just getting this PR in?

PR #145, opened 2022-06-06
Add pypy ci

dannyjacobs:
This repo is now installable from pipy. This is the preferred installation method for normal users and testing. This PR also adds a github action which causes tagged versions to be built and uploaded to pipy.

### DIFF
--- a/.github/workflows/publish-to-test-pypi.yaml
+++ b/.github/workflows/publish-to-test-pypi.yaml
@@ -1,0 +1,49 @@
+name: Publish Python distributions to PyPI and TestPyPI
+
+on: [push, pull_request]
+
+jobs:
+  build-n-publish:
+    name: Build and publish to PyPI and TestPyPI
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        # Adding -l {0} helps ensure conda can be found properly.
+        shell: bash -l {0}
+    env:
+      ENV_NAME: publish
+      PYTHON: 3.8
+    steps:
+      - uses: actions/checkout@master
+        with:
+          fetch-depth: 0
+
+      - name: Setup Miniconda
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-to-test-pypi.yaml
```diff
@@ -0,0 +1,49 @@
...
+          conda list
+          PYVER=`python -c "import sys; print('{:d}.{:d}'.format(sys.version_info.major, sys.version_info.minor))"`
+          if [[ $PYVER != ${{ env.PYTHON }} ]]; then
+            exit 1;
+          fi
+
+
+      - name: Build a binary wheel and a source tarball
+        run: |
+          python -m pep517.build --source --out-dir dist/ .
```
steven-murray, line 42: apparently you should just use `python -m build .` these days

.github/workflows/publish-to-test-pypi.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+
+      - name: Publish to PyPI
+        if: startsWith(github.event.ref, 'refs/tags')
+        uses: pypa/gh-action-pypi-publish@master
```
steven-murray, line 47: is it still master, or is it main?

steven-murray (commented):
Almost.

.github/workflows/publish-to-test-pypi.yaml
bhazelton: It's master.

.github/workflows/publish-to-test-pypi.yaml
dannyjacobs: done

bhazelton (approved):
seems reasonable to me.

PR #149, opened 2022-08-15
Add plot_antclass function for plotting `hera_qm.ant_class.AntennaClassification` objects

jsdillon:
This PR adds a new plotting routine for use in the file_inspect notebook and anywhere else that uses `AntennaClassification` objects.


jsdillon:
Here's an example use:

image

tyler-a-cox (approved):
Looks good, thanks @jsdillon

PR #151, opened 2022-12-30
fix: remove deprecated numpy types

steven-murray:
Removes all the `np.int`, `np.float`, `np.complex` and `np.str`


jsdillon (approved):
Looks good to me

PR #152, opened 2023-01-04
Better Packaging

steven-murray:
This adds a few nicer packaging / CI goodies, like `setuptools_scm` (like the rest of the stack), a simpler `pyproject.toml` packaging config, auto-labeling of PRs, auto release-drafting, easier testing and better flake8 configuration. 

Will add `black` in another PR.


README.md
```diff
@@ -26,3 +26,11 @@ tests. All of these packages can be installed with the following commands::
 To test the package, execute the following command::
 
     $ nosetests uvtools/tests/test_dspec.py uvtools/tests/test_utils.py
+
+## Making a Release
+
+To make a release of `aipy` (both on Github and PyPI), head to the most current
```
mkolopanis, line 32: I think we're suffering from copy-pasta here

pyproject.toml
```diff
@@ -0,0 +1,44 @@
+[build-system]
+requires = ["setuptools", "setuptools-scm[toml]>=6.2", "numpy"]
```
mkolopanis, line 2: we need numpy for build? if so we might consider changing this to `"oldest-supported-numpy"` depending on the need. the oldest-supported is for api compat if I remember correctly but it's been a while since I delved into it deeply.

pyproject.toml
```diff
@@ -0,0 +1,44 @@
...
+    "pyuvdata",
+    "astropy>5.1"
+]
+dynamic = [
+    "version",
+]
+
+
+[project.optional-dependencies]
+aipy = ["aipy>=3.0rc2"]
```
mkolopanis, line 31: should `aipy` be listed under dependencies _and_ optional?

mkolopanis (commented):
A few quick questions

pyproject.toml
steven-murray: Good call. It's optional.

pyproject.toml
steven-murray: nope, I added that to test stuff out because aipy wasn't building properly, but I think it should work without it.

README.md
steven-murray: Aye.

steven-murray:
@mkolopanis thanks for the review! I've fixed all those things and tests are back to passing :-)

mkolopanis (approved):
thanks for the updates, looks good to me

PR #153, opened 2023-01-10
ci: add config files for auto-labelling and release drafting

steven-murray:
This is the first step in adding better packaging/release mechanisms to uvtools (see #152). This has to be merged to main for some of the workflows in that PR to work. After everything is in, releasing should follow the new way implemented in https://github.com/hera-team/aipy. In brief:

* On each push/PR, we automatically create (or update) a Draft Github Release with the unreleased changes. This serves as an automatic online changelog, rather than spotty manual updating of a text changelog. This is just a DRAFT.
* Also on each push, we check that the distribution builds and that `twine check` runs without error (but we don't release to PyPI). 
* The Draft Release has an automatically *suggested* version, based on labels applied to all un-released merged PRs. These labels are semi-automatically applied based on rules in the configuration defined in this PR (in the release-drafter yaml). Of course, manual labels can also be applied. 
* When a new release is desired, a maintainer simply has to go to the newest draft release, which has the suggested version already. They can modify this version if they like, but mostly everything should already be in-tact. They just have to hit the "Publish" button. This sets off another workflow which uploads the package to PyPI with the latest tag (which corresponds to the release). 

I think this is an easy and very visual way to do releases. I noticed that uvtools in particular has a poorly maintained changelog, so some of this automation is probably very useful.

### DIFF
--- a/.github/labels.yml
+++ b/.github/labels.yml
@@ -43,9 +43,6 @@
 - name: performance
   description: Performance
   color: "016175"
-- name: python
-  description: Pull requests that update Python code
-  color: 2b67c6
 - name: question
   description: Further information is requested
   color: d876e3
--- a/.github/labels.yml
+++ b/.github/labels.yml
@@ -1,0 +1,66 @@
+---
+# Labels names are important as they are used by Release Drafter to decide
+# regarding where to record them in changelog or if to skip them.
+#
+# The repository labels will be automatically configured using this file and
+# the GitHub Action https://github.com/marketplace/actions/github-labeler.
+- name: breaking
+  description: Breaking Changes
+  color: bfd4f2
+- name: bug
+  description: Something isn't working
+  color: d73a4a
+- name: build
+  description: Build System and Dependencies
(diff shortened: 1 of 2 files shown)


.github/labels.yml
```diff
@@ -0,0 +1,66 @@
...
+  color: 008672
+- name: invalid
+  description: This doesn't seem right
+  color: e4e669
+- name: performance
+  description: Performance
+  color: "016175"
+- name: python
+  description: Pull requests that update Python code
+  color: 2b67c6
```
bhazelton, line 48: This seems so broad that it might be useless. And it doesn't look like it's being used by the auto labeler, which seems like it would be ideally placed to determine if any python code had been modified.

bhazelton (approved):
Seems reasonable and will enable the big packaging PR. There's one label I commented on but feel free to ignore.

.github/labels.yml
steven-murray: Good point! Removed.

